PR #197, opened 2024-03-28
add verify prod and refactor terraform check/deploy

### LINKED ISSUE #54: Setup a "plan" presubmit
This needs two separate PRS.

1. We should create a NEW identity in `iac/bootstrap` that's a project viewer and can be federated by pull requests presubmit (e.g. `ref:refs/heads/main` -> `:pull_request`). -- I may need to apply this once it lands, or add folks to ACLs.

1. We should add a presubmit like mono that posts the plan to the PR body.

cc @cpanato

cpanato:
Fixes #54

Need https://github.com/octo-sts/app/pull/196

### DIFF
--- a/.github/workflows/.terraform.yaml
+++ b/.github/workflows/.terraform.yaml
@@ -1,0 +1,216 @@
+# Copyright 2024 Chainguard, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Reusable Terraform Workflow
+
+on:
+  workflow_call:
+    inputs:
+      workload_identity_provider:
+        type: string
+        required: true
+        description: "Workload identity provider for GCP Workload Federation."
+      service_account:
+        type: string
+        required: true
+        description: "GCP Service Account to impersonate."
+      checkout_sha:
+        required: false
+        type: string
+        description: 'Commit SHA to run at.'
+      working_directory:
+        required: true
+        type: string
+        description: 'Location for the terraform.'
+      project_id:
+        required: false
+        type: string
+        description: 'GCP project ID.'
+      octo-sts-identity:
+        required: false
+        default: ''
+        type: string
+        description: 'The Octo STS identity name'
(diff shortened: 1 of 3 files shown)


.github/workflows/verify-prod.yaml
```diff
@@ -0,0 +1,29 @@
+# Copyright 2024 Chainguard, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Verify prod Octo-sts
+
+on:
+  pull_request_target:
+    branches:
+      - 'main'
```
mattmoor, line 9: I think we want to add a labeling requirement here as well, since this repo is now public.

cpanato: oh, ok, so we will need to trigger when is updated as well

cpanato: are we doing a similar thing in another repo? or will this be the first one?

mattmoor: we should, but not yet

.github/workflows/.terraform.yaml
```diff
@@ -0,0 +1,216 @@
...
+
+      - name: Terraform init
+        id: init
+        run: terraform init -reconfigure
+
+      - name: Terraform Validate
+        id: validate
+        run: terraform validate -no-color
+
+      - name: Terraform Plan
```
mattmoor, line 103: Do you know of anything that we can run over this to check for injection vulnerabilities?

Skimming, the only one that stood out is `inputs.working_directory`, and that's a reach, but would be good to sanity check things

mattmoor (approved):

mattmoor:
I'm going to merge this, so that we can start using it (probably would have avoided mild outage yesterday).  I'd love an answer for the shell check issue.

SamyDgz (approved):

PR #247, opened 2024-05-06
Add webhook to validate STS configurations at pull request.

wlynch:
Since we don't give any information at request time about why a STS policy is valid or not so that we don't leak details about the policy, this webhook will validate on the repo at pull request / push so the repo authors get feedback on potential config issues.

This changes does not set up GCLB, since this probably needs some more discussion for how we set this up to not conflict with the existing STS service.


mattmoor:
Hey @wlynch, sorry for neglecting this.  I wanted to get an image out for the app, and refactor the TF a bit to make deploying this to dev environments more tractable.

I'm going to TAL at this now, and try to refactor this on top of https://github.com/octo-sts/app/pull/284 to get this up in my dev environment 🤞

mattmoor:
When I make edits in `mattmoor/test-bed` I'm seeing `check_suite.requested` events, which aren't handled.  I'm gonna futz around a bit (it looks like we handle this in source enforce) 🤞 

ref: https://github.com/mattmoor/test-bed/pull/1

Issue #288, opened 2024-05-20
Errors during `check_suite.requested`


mattmoor:
I disabled the webhook for now because we are seeing errors like:
```
ERROR error handling event *github.CheckSuiteEvent: GET https://api.github.com/repos/chainguard-dev/.github/compare/0000000000000000000000000000000000000000...5d7c5a22d4828b2576a08aa4e5cd70ca68d4579e: 404 Not Found [] X-GitHub-Delivery=b2c255e0-16ca-11ef-81e6-834ee7dc378d X-GitHub-Event=check_suite
```

This may be my own user error trying to handle the check suite events, but we should sort it out before a 0.2.0.

cc @wlynch @cpanato

wlynch:
The zero commit SHA is a convention to represent nil - it's not a reference-able object id.

e.g.

```sh
$ gh api /repos/chainguard-dev/.github/commits/0000000000000000000000000000000000000000
{
gh: No commit found for SHA: 0000000000000000000000000000000000000000 (HTTP 422)
  "message": "No commit found for SHA: 0000000000000000000000000000000000000000",
  "documentation_url": "https://docs.github.com/rest/commits/commits#get-a-commit"
}
```

mattmoor:
Yeah, totally makes sense.  Need to figure out how we want to deal with that case properly in the `check_suite` events.

Not a rush, but happy to chat about it if you have cycles sometime.

cpanato:
Is this still an issue?

mattmoor:
Yeah, I don’t think I have re-enabled things because of this

PR #309, opened 2024-05-30
if we got a zerohash get the contents of the directory that we are interested

cpanato:
Fixes: #288 

not sure if that is correct, but was my understanding based on the slack discussion

cc @wlynch


cpanato:
this PR https://github.com/octo-sts/app/pull/308 should fix the failing ci check

wlynch (approved):

wlynch (approved):

PR #412, opened 2024-07-25
feat: allow flexible options for github app secret and metrics

karlhaworth:
feat: allow flexible options for github app secret and metrics

allows for the GitHub app secret to be loaded from GCPKMS as the default, but also now allows envvars and files to be utilized. Vaulting mechanisms are used in our case, however they're loaded a bit differently.

allows for the metrics specific to chainguard to be optionally disabled. the default is to allow the chainguard default metrics.

### DIFF
--- a/pkg/envconfig/envconfig.go
+++ b/pkg/envconfig/envconfig.go
@@ -8,5 +8,5 @@ type EnvConfig struct {
 	EventingIngress            string `envconfig:"EVENT_INGRESS_URI" required:"true"`
 	AppSecretCertificateFile   string `envconfig:"APP_SECRET_CERTIFICATE_FILE" required:"false"`
 	AppSecretCertificateEnvVar string `envconfig:"APP_SECRET_CERTIFICATE_ENV_VAR" required:"false"`
-	Metrics                    string `envconfig:"METRICS" required:"false" default:"default"`
+	Metrics                    bool   `envconfig:"METRICS" required:"false" default:"true"`
 }
--- a/pkg/envconfig/envconfig.go
+++ b/pkg/envconfig/envconfig.go
@@ -1,0 +1,12 @@
+package envconfig
+
+type EnvConfig struct {
+	Port                       int    `envconfig:"PORT" required:"true"`
+	Domain                     string `envconfig:"STS_DOMAIN" required:"true"`
+	KMSKey                     string `envconfig:"KMS_KEY" required:"false"`
+	AppID                      int64  `envconfig:"GITHUB_APP_ID" required:"true"`
+	EventingIngress            string `envconfig:"EVENT_INGRESS_URI" required:"true"`
+	AppSecretCertificateFile   string `envconfig:"APP_SECRET_CERTIFICATE_FILE" required:"false"`
+	AppSecretCertificateEnvVar string `envconfig:"APP_SECRET_CERTIFICATE_ENV_VAR" required:"false"`
+	Metrics                    string `envconfig:"METRICS" required:"false" default:"default"`
+}
--- a/pkg/envconfig/envconfig.go
+++ b/pkg/envconfig/envconfig.go
@@ -1,5 +1,11 @@
 package envconfig
 
+import (
+	"errors"
+
+	"github.com/kelseyhightower/envconfig"
+)
+
 type EnvConfig struct {
 	Port                       int    `envconfig:"PORT" required:"true"`
 	Domain                     string `envconfig:"STS_DOMAIN" required:"true"`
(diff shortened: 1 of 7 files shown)


karlhaworth:
@wlynch let me know if you have feedback or questions. Thanks!

pkg/envconfig/envconfig.go
```diff
@@ -0,0 +1,12 @@
...
+type EnvConfig struct {
+	Port                       int    `envconfig:"PORT" required:"true"`
+	Domain                     string `envconfig:"STS_DOMAIN" required:"true"`
+	KMSKey                     string `envconfig:"KMS_KEY" required:"false"`
+	AppID                      int64  `envconfig:"GITHUB_APP_ID" required:"true"`
+	EventingIngress            string `envconfig:"EVENT_INGRESS_URI" required:"true"`
+	AppSecretCertificateFile   string `envconfig:"APP_SECRET_CERTIFICATE_FILE" required:"false"`
+	AppSecretCertificateEnvVar string `envconfig:"APP_SECRET_CERTIFICATE_ENV_VAR" required:"false"`
+	Metrics                    bool   `envconfig:"METRICS" required:"false" default:"true"`
+}
```
wlynch, line 12: It's probably worth adding some validation to return an error if multiple KMS options are set.

e.g. 

```go
func Process() (*EnvConfig, error) {
  cfg := new(EnvConfig)
  if err := envconfig.Process("", cfg); err != nil {
    return nil, err
  }

  kmsSet := false
  for _, v := range []string{cfg.KMSKey, cfg,AppSecretCertificateFile, cfg.AppSecretCertificateEnvVar} {
    if v != "" {
      if kmsSet {
        return nil, errors.New("only one of KMS_KEY, APP_SECRET_CERTIFICATE_FILE, APP_SECRET_CERTIFICATE_ENV_VAR may be set")
      }
      kmsSet = true
    }
  }
  
  return cfg, err

File: pkg/ghtransport/ghtransport_test.go [hunk 2]
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	privateKeyBytes := x509.MarshalPKCS1PrivateKey(privateKey)
+
+	privateKeyPEM := pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: privateKeyBytes,
+	}
+
+	tmpFile, err := os.CreateTemp("", "privateKey*.pem")
```
wlynch, line 160:
```suggestion
	tmpFile, err := os.CreateTemp(t.TempDir(), "privateKey*.pem")
```

pkg/ghtransport/ghtransport_test.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return client
+}
+
+func createGCPKMSCredsFile(t *testing.T) string {
+
+	tmpFile, err := os.CreateTemp("", "creds-")
```
wlynch, line 109:
```suggestion
	tmpFile, err := os.CreateTemp(t.TempDir(), "creds-")
```

I know you have an os.Remove above, but this is generally a safer way to ensure the temp file is always cleaned up after the test runs.

pkg/ghtransport/ghtransport_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+func TestGCPKMS(t *testing.T) {
+	ctx := context.Background()
+
+	credsFile := createGCPKMSCredsFile(t)
+
+	defer os.Remove(credsFile)
+
+	os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", credsFile)
+	defer os.Unsetenv("GOOGLE_APPLICATION_CREDENTIALS")
```
wlynch, line 32:
```suggestion
	t.Setenv("GOOGLE_APPLICATION_CREDENTIALS", credsFile)
```

wlynch (commented):
Looks good! Suggested some small tweaks.

cc @mattmoor in case you have any concerns.

pkg/envconfig/envconfig.go
karlhaworth: [878994d](https://github.com/octo-sts/app/pull/412/commits/878994d0e542db889818fb31943927dcbd2f8c29) should do it. thanks

karlhaworth:
fixed all `golangci-lint run` reports and boilerplate reports after the workflows were approved. Thank you.

pkg/ghtransport/ghtransport.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/octo-sts/app/pkg/gcpkms"
+)
+
+func New(ctx context.Context, env *envConfig.EnvConfig, kmsClient *kms.KeyManagementClient) (*ghinstallation.AppsTransport, error) {
+	switch {
+	case env.AppSecretCertificateEnvVar != "":
+		atr, err := ghinstallation.NewAppsTransport(http.DefaultTransport, env.AppID, []byte(env.AppSecretCertificateEnvVar))
+
+		if err != nil {
+			log.Panicf("error creating GitHub App transport: %v", err)
```
wlynch, line 23: return err instead of panic

wlynch (changes requested):
one last change, otherwise LGTM! 🎉 

Thanks!

pkg/ghtransport/ghtransport.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+		return atr, nil
+	default:
+		if env.KMSKey == "" {
+			log.Panicf("failed to process env var: %s", env.KMSKey)
+		}
+
+		signer, err := gcpkms.New(ctx, kmsClient, env.KMSKey)
+		if err != nil {
+			log.Panicf("error creating signer: %v", err)
```
wlynch, line 41:
```suggestion
			return nil, fmt.Errorf("failed to process env var: %q", env.KMSKey)
		}

		signer, err := gcpkms.New(ctx, kmsClient, env.KMSKey)
		if err != nil {
			return nil, fmt.Errorf("error creating signer: %w", err)
```

wlynch (approved):

wlynch (approved):

karlhaworth:
@wlynch any chance we could create a release as well so I can use the image from https://images.chainguard.dev/directory/image/octo-sts/versions? Not sure how that image gets updated. I appreciate your time!

wlynch:
Yes! we can cut a release shortly. I'll work on getting the dependabot updates in first though.

PR #449, opened 2024-08-12
Add tokensource library

wlynch:
Adds library to make it easy to create a oauth2.TokenSource for use with GitHub clients.

### DIFF
--- a/pkg/tokensource/tokensource.go
+++ b/pkg/tokensource/tokensource.go
@@ -1,0 +1,116 @@
+// Copyright 2024 Chainguard, Inc.
+// SPDX-License-Identifier: Apache-2.0
+package tokensource
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	"chainguard.dev/sdk/sts"
+	"github.com/chainguard-dev/clog"
+	"golang.org/x/oauth2"
+	"golang.org/x/time/rate"
+)
+
+type TokenSource struct {
+	ctx                   context.Context
+	url                   string
+	org, repo, policyName string
+	// Base token source (e.g. the cred we send to the STS service to exchange).
+	base      oauth2.TokenSource
+	sometimes rate.Sometimes
+
+	// Output fields.
+	tok *oauth2.Token
+	err error
+}
+
+func NewTokenSource(ctx context.Context, stsURL string, org, repo, policyName string, ts oauth2.TokenSource) *TokenSource {
+	return &TokenSource{
+		ctx:        ctx,
+		url:        stsURL,
+		org:        org,
+		repo:       repo,
+		policyName: policyName,
+		base:       ts,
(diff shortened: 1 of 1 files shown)


imjasonh:
Actually I'd probably not use a library that pulls in k8s deps just to get this in tf-common. Plus it depends on tf-common itself 😆 

Any chance we could make an octo-sts/client repo? 🤷

wlynch:
I'm for that. 

@mattmoor objections?

mattmoor:
I'm missing something, what's the K8s dep?

wlynch:
I think jason is referring to the deps in go.mod for the server components. IIIUC even though the Go tool is smart enough to not include them in produced binaries if they're unused, they'll still show up in go.mod of dependencies as indirect and you'll need to dependabot bump them regularly.

mattmoor:
I see a single line in the entire repo using the API machinery `sets` which could simply use something else, and I am not sure that `sigs.k8s.io/yaml` is anything to be worried about (I believe it's a fork of another yaml package to support `json` annotations, so I think it's well scoped/contained).

jmeridth:
:robot: (critical) AI-assisted review -- 2 models (Opus 4.8, Fable 5).

Closing this as stale/superseded -- the idea is good but it needs a fresh implementation rather than a rebase. Summary of what we found:

**Won't build against current main.** The branch pins `chainguard.dev/sdk v0.1.19` where `sts.Exchange` returned `(string, error)`; main is on `v0.1.156` where it returns `(sts.TokenPair, error)`. `return res` no longer compiles -- it'd need `res.AccessToken`.

**Deadlocks on the first token refresh.** `rate.Sometimes.Do` holds its mutex for the whole callback (its own docs warn that re-entrancy deadlocks). The refresh callback calls `Revoke()` -> `oauth2.NewClient(ctx, ts).Do(req)`, and the oauth2 transport synchronously calls `ts.Token()`, which re-enters `sometimes.Do` on the held mutex. It works for the first ~45 min, then the first successful refresh hangs the source permanently.

**Org-only scope is silently wrong.** `scope = fmt.Sprintf("%s/%s", org, repo)` yields `"org/"` when `repo` is empty; the octo-sts server parses that as a repo literally named `org` instead of the org-level `.github` policy. Should only append `/repo` when `repo != ""`.

Also non-blocking: data race on the unlocked `return ts.tok, ts.err`, a failed refresh poisoning the source for 45 min (and dropping the still-valid old token unrevoked), and the swallowed `Revoke()` error.

Separately, the original thread left open whether this belongs here or in a dedicated `octo-sts/client` repo -- worth settling before a rewrite. FWIW the k8s-dep concern from 2024 doesn't apply to hosting it in this module (those deps are already present; this only promotes `golang.org/x/time` to a direct dep). Thanks @wlynch -- happy to help land a v2 if it's still wanted.

Issue #469, opened 2024-08-28
Policy Definitions for STS Policies


karlhaworth:
# Policy Definitions for STS Policies

**Outcome** - Deny undesired cross-org access

@wlynch mentioned possibly incorporating into -> https://github.com/octo-sts/app/blob/main/pkg/webhook/webhook.go

## Problem

In large orgs, if someone were able to commit a file with an undesired org, this undesired org has access to a private org.

As a octo-sts implementor,
I want to restrict the orgs that can obtain tokens in octo-sts
So that I can be assured no unauthorized org can ever gain access using octo-sts

## Potential Solution

Envisioning a way to instruct Octo-STS to only hand out tokens to valid declared orgs in a policy.

In the STS policies there's already the `subject_pattern` being used. I am proposing to use the org bounded claim instead of repo if possible for this.

The policy definition would live at the org level under something like `.github/.chainguard/policy.sts.yaml`.

```yaml
kind: policy

issuer: https://token.actions.githubusercontent.com
subject_patterns:
  - org:AATestOrg
```

figured maybe we'd need a `kind` to be added?

karlhaworth:
@wlynch 

what about something like the below in `pkg/octosts/octosts.go`? The `ALLOWED_ORGS` can be a comma separated list. Similar to what I saw here https://github.com/octo-sts/app/pull/476/files#diff-3e455a0e8a009c415c69c239ee08525a8f1ae4c0a127fdd94297ec0b1ef262ce

real quick stab at this.

```
	parsedToken, err := jwt.Parse(bearer, nil)
	if parsedToken == nil {
		return nil, err
	}
	claims, _ := parsedToken.Claims.(jwt.MapClaims)

	orgs := strings.Split(s.orgs, ",")

	clog.FromContext(ctx).Infof("claims: %#v", claims)

	repositoryOwner := claims["repository_owner"].(string)
	isValidOwner := false
	for _, org := range orgs {
		if org == repositoryOwner {
			isValidOwner = true
			break
		}
	}

	if !isValidOwner {
		return nil, status.Errorf(codes.PermissionDenied, "invalid repository owner")
	}
```

wlynch:
Some notes that were lost in slack that I wanted to reshare here for visibility.

[...]

PR #476, opened 2024-08-30
Add org filter to webhook.

wlynch:
We need to test out behavior in the webhook, but we're not 100% confident it's working as intended as is. This adds a configurable filter to enable selective responses to certain orgs.

For now, have the default deployment only look at octo-sts and chainguard-dev orgs.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -41,6 +41,8 @@ type Validator struct {
 	// Store multiple secrets to allow for rolling updates.
 	// Only one needs to match for the event to be considered valid.
 	WebhookSecret [][]byte
+
+	Organizations []string
 }
 
 func (e *Validator) ServeHTTP(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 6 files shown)


karlhaworth:
Might be missing something here, but let's say the web hook doesn't stop the config from being merged. What in this PR is stopping the app from handing out the token anyways to a non-specified org?

wlynch:
Great question! It doesn't 😅 This is not related to https://github.com/octo-sts/app/issues/469.

The webhook is currently disabled on webhook.octo-sts.dev because of https://github.com/octo-sts/app/issues/288.
We want to re-enable it, but we want to be careful not to start spamming everyone who has the app installed currently. This change effectively adds a preview flag to only have the webhook respond to certain orgs so we can make the fixes we need and validate it in specific orgs before we drop the flag and enable it for everyone.

pkg/webhook/webhook.go
```diff
@@ -256,6 +264,12 @@ func (e *Validator) handlePullRequest(ctx context.Context, pr *github.PullReques
 	sha := pr.GetPullRequest().GetHead().GetSHA()
 	installationID := pr.GetInstallation().GetID()
 
+	// Skip if the organization is not in the list of organizations to validate.
+	if e.shouldSkipOrganization(owner) {
```
cpanato, line 268: should we normalize to all lowercase in case the operator set a organization with lowercases and what we get is not, like

in the config: `myorg`
in the request `MyOrg`

cpanato (dismissed):
small nit comment, lgtm

thanks

pkg/webhook/webhook.go
wlynch: Done.

cpanato (approved):
cool

PR #486, opened 2024-09-03
Fix yaml tag parsing

wlynch:
Switches yaml parsing library to sigs.k8s.io/yaml which respects json field tags. This fixes yaml parsing for policies, since previously it was not respecting the tags.

Also adds a test!

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -20,8 +20,8 @@ import (
 	"github.com/google/go-github/v61/github"
 	"github.com/hashicorp/go-multierror"
 	"github.com/octo-sts/app/pkg/octosts"
-	"gopkg.in/yaml.v2"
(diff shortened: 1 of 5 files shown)


pkg/webhook/webhook.go
```diff
@@ -20,8 +20,8 @@ import (
 	"github.com/google/go-github/v61/github"
 	"github.com/hashicorp/go-multierror"
 	"github.com/octo-sts/app/pkg/octosts"
-	"gopkg.in/yaml.v2"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/yaml"
```
cpanato, line 24: :)

cpanato (approved):

pkg/webhook/webhook.go
wlynch: it's my favorite

PR #508, opened 2024-09-15
Support for custom audience

### LINKED ISSUE #259: Add support for customizing acceptable `aud`
Today you can't really customize the `aud` from most SSO flows.

Today this is checked here:
https://github.com/octo-sts/app/blob/1fc549c0973c27a9aba0257626d1903486dee2d9/pkg/octosts/octosts.go#L116-L118

We could check this here (and default to `octo-sts.dev`):
https://github.com/octo-sts/app/blob/1fc549c0973c27a9aba0257626d1903486dee2d9/pkg/octosts/trust_policy.go#L131-L134

pdeslaur:
Fixes #259

Adds the option for a trust policy to specify which audience is expected from the upstream token.

- Add support for the `audience` and `audience_pattern`: The trust policy is accepted if at least one of the provided audience matches. 
- Adds support for boolean claims: If a boolean claim is found in the upstream token, it is converted to the "true" or "false" string

### DIFF
--- a/pkg/octosts/trust_policy_test.go
+++ b/pkg/octosts/trust_policy_test.go
@@ -4,7 +4,9 @@
 package octosts
 
 import (
+	"reflect"
 	"testing"
+	"unsafe"
 
 	"github.com/coreos/go-oidc/v3/oidc"
 )
@@ -107,6 +109,7 @@ func TestCheckToken(t *testing.T) {
 		name    string
 		tp      *TrustPolicy
 		token   *oidc.IDToken
+		claims  []byte
 		wantErr bool
 	}{{
 		name: "valid token",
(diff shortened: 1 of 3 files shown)


pkg/octosts/trust_policy_test.go
```diff
@@ -199,10 +241,48 @@ func TestCheckToken(t *testing.T) {
...
+		name: "support boolean claims",
+		tp: &TrustPolicy{
+			Issuer:  "https://example.com",
+			Subject: "blah",
+		},
+		token: &oidc.IDToken{
+			Issuer:  "https://example.com",
+			Subject: "blah",
+		},
+		wantErr: false,
```
mattmoor, line 285: Is this missing something?

pkg/octosts/trust_policy_test.go
```diff
@@ -199,10 +241,48 @@ func TestCheckToken(t *testing.T) {
...
 	}}
 
 	// TODO(mattmoor): Figure out how to test custom claims with IDToken.
```
mattmoor, line 288: There's a reflection trick I found after we initially built this, which should make this possible!

Let me dig it up

mattmoor: https://github.com/chainguard-dev/mono/blob/6856b8058b0a43170f5a49336c97f2fbf0943f1e/registry/internal/fake/token.go#L92-L100

pkg/octosts/trust_policy_test.go
pdeslaur: Fixed!

pkg/octosts/trust_policy_test.go
pdeslaur: Neat and gross! Added a bunch of tests to cover the list - lmk what you think

mattmoor (approved):

PR #583, opened 2024-11-04
feat: parent trust policy at org root

karlhaworth:
feat: parent trust policy at org root

ensures that users within an org cannot commit unsafe trust files according to enterprise policies.

Another try at https://github.com/octo-sts/app/issues/469.

Similar to org wide trust policies living at https://github.com/chainguard-dev/.github/tree/main/.github/chainguard, this would allow a parent level trust policy at the org root.

`https://github.com/TestOrg/.github/blob/master/.github/chainguard/org.sts.yaml`

Content of `.github/chainguard/org.sts.yaml`

```
issuer: https://token.actions.githubusercontent.com
subject_pattern: repo:TestOrg/.*:ref:refs/.*
```

Logic could be improved to work without the `ORG_TRUST_POLICY` env var. I introduced this to force the `ORG_TRUST_POLICY` behavior.

Simple approach, but happy for feedback either way.

### DIFF
--- a/cmd/app/main.go
+++ b/cmd/app/main.go
@@ -37,6 +37,10 @@ func main() {
 		log.Panicf("failed to process env var: %s", err)
 	}
 
+	if baseCfg.OrgTrustPolicy {
+		clog.FromContext(ctx).Infof("Required Org Policy Active")
+	}
+
 	if baseCfg.Metrics {
 		go metrics.ServeMetrics()
 
--- a/cmd/app/main.go
+++ b/cmd/app/main.go
@@ -71,7 +71,7 @@ func main() {
 		log.Panicf("failed to create cloudevents client: %v", err)
 	}
 
-	pboidc.RegisterSecurityTokenServiceServer(d.Server, octosts.NewSecurityTokenServiceServer(atr, ceclient, appConfig.Domain, baseCfg.Metrics))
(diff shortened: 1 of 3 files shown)


jmeridth:
:robot: (low, non-blocking) AI-assisted review.

Recommending we close this as stale/superseded -- the surface it patches has been refactored on main, so it no longer compiles or merges:

- `NewSecurityTokenServiceServer` is now `(rrm ghinstall.Manager, sticky stickystore.Store, appCount int, ceclient, domain, metrics)` -- the single-`atr` constructor and `sts` struct this edits no longer exist.
- `lookupInstallAndTrustPolicy` is now `(ctx, scope, identity, subject) -> (*AppsTransport, int64, *OrgTrustPolicy, error)`; the call here passes 3 args expecting 3 returns.
- Org-level `.github` policy resolution already landed on main (`lookupInstallAndTrustPolicy` handles `repo == ".github"`).

The defense-in-depth idea here -- requiring every exchange to also satisfy an org-root policy -- is still worthwhile, but it would need a fresh implementation against current main and should be coordinated with the active org-policy work in #1283 (`ORG_POLICY_REPO`) and #901 (org trusted issuers) so we don't end up with three overlapping mechanisms. One implementation note for a v2: clone the request with `proto.Clone` rather than `encoding/json`, and drop the `%#v` request/policy logging at `Info`.

Thanks @karlhaworth -- closing this one, but happy to review a rebased take if you want to pick it back up.

PR #586, opened 2024-11-05
webhook: check for non-.sts.yaml files in `.github/chainguard`

imjasonh:
An easy way to mess up Octo STS is to create a policy file named like `.github/chainguard/my-policy.yaml` (ask me how I know! 😆)

Webhook validation currently only validates the structure of valid policy file paths, but it will happily let you commit non-STS file paths, which cause problems later when you try to use `my-policy`.

This may break existing repos using Octo STS, but I'd guess that if they have a `my-policy.yaml` they won't also have a valid `my-policy.sts.yaml`, so any breakage would probably be highlighting a real issue. In any case, if that's a problem, we can make this a warning instead of a failing check, if we want to.


pkg/webhook/testdata/api/v3/repos/foo/bar/compare/1234...2345
```diff
@@ -0,0 +1,422 @@
+{
```
imjasonh, line 1: This is a direct copy of `1234...5678` with all `test.sts.yaml`s replaced with `test.yaml`.

pkg/webhook/webhook.go
```diff
@@ -247,10 +246,19 @@ func (e *Validator) handlePush(ctx context.Context, event *github.PushEvent) (*g
...
 	}
+	var nonSTSFiles []string
+	for _, f := range files {
+		if !strings.HasSuffix(f, ".sts.yaml") {
+			nonSTSFiles = append(nonSTSFiles, f)
+		}
+	}
+	if len(nonSTSFiles) > 0 {
+		return e.handleNonSTSFiles(ctx, client, owner, repo, sha, nonSTSFiles)
+	}
```
mattmoor, line 261: If there are non-sts files we ignore all other validation?

pkg/webhook/webhook.go
```diff
@@ -411,3 +439,25 @@ func (e *Validator) shouldSkipOrganization(org string) bool {
...
+		Name:        "Trust Policy Validation",
+		HeadSHA:     sha,
+		ExternalID:  github.String(sha),
+		Status:      github.String("completed"),
+		Conclusion:  github.String("failure"),
+		StartedAt:   &github.Timestamp{Time: time.Now()},
+		CompletedAt: &github.Timestamp{Time: time.Now()},
+		Output: &github.CheckRunOutput{
+			Title:   github.String("Non-STS YAML file(s)."),
+			Summary: github.String("Found non-STS YAML files in `.github/chainguard` directory:\n\n" + strings.Join(nonSTSFiles, "\n")),
```
mattmoor, line 455: I don't see where you check for `.yaml`?

This would flag `.github/chainguard/README.md` IIUC

imjasonh: Good catch, fixed.

pkg/webhook/webhook.go
imjasonh: That's true, yeah. I didn't want to reach too far into the "real" validation method to account for the case where there's both kinds of errors, but I can if you'd prefer that.

With this change as it is, if someone had an invalid `.sts.yaml` and a non-sts-`.yaml` they'd get a failure for the latter, fix it, and get another failure for the former, which is kind of sub-optimal, but at least it doesn't let bad stuff sneak through.

pkg/webhook/webhook.go
```diff
@@ -378,11 +395,22 @@ func (e *Validator) handleCheckSuite(ctx context.Context, cs checkSuite) (*githu
 			return nil, err
 		}
 		for _, file := range resp {
-			if ok, err := filepath.Match(".github/chainguard/*.sts.yaml", file.GetFilename()); err == nil && ok {
+			if strings.HasPrefix(file.GetFilename(), ".github/chainguard/") && strings.HasSuffix(file.GetFilename(), ".yaml") {
```
cpanato, line 398: how about the `.yml` ?

cpanato: i can do that in a follow up

cpanato (dismissed):
lgtm

mepi262:
@imjasonh 
It seems that this pull request has conflicts.
Would you resolve it?

jdolitsky (approved):

tcnghia (dismissed):

EyeCantCU (approved):

jmeridth:
:robot: (low, non-blocking) AI-assisted review.

Closing this as stale -- it was approved back in Nov 2024 but never rebased, and `webhook.go` has been substantially rewritten since (#1379, #1403, #1404, #1411, #1412, #1437), so it's conflicting and the file structure it patches is gone. The feature itself (warning on non-`.sts.yaml` files in `.github/chainguard`) still isn't on main and is still worth having.

If someone picks it back up, a fresh implementation against current `webhook.go` would be cleaner than untangling the conflicts, and it'd be a good chance to fold in the two threads left open here: mattmoor's point that a stray file currently short-circuits the rest of validation, and cpanato's `.yml` / other-extension follow-up. Closing for now to keep the queue honest -- easy to reopen or resubmit.

PR #587, opened 2024-11-05
Add option to prefer fetching policies from org repo.

wlynch:
This adds the ability to have a server-side flag to allow configuring OctoSTS to only read policies from the `org/.github` repo. This allows for policy centralization for deployments that prefer stricter consolidation and enforcement.

This also adds a test harness for the exchange flow to make it easier to test and see what values are being exchanged.

This is a alternative to https://github.com/octo-sts/app/pull/583. This PR aims to provide similar functionality, but preserve incoming identity + scope details. cc @karlhaworth


wlynch:
Not proceeding with this - will look into modifying https://github.com/octo-sts/app/pull/583

PR #745, opened 2025-02-07
handle and reply when accessing /

cpanato:
not sure if is the best place or that is the way to handle that, but trying to reply a better message when people access https://octo-sts.dev/ 

```
{"code":5, "message":"Not Found", "details":[]}
```

### DIFF
--- a/cmd/app/main.go
+++ b/cmd/app/main.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"log"
 	"log/slog"
+	"net/http"
 	"os"
 	"os/signal"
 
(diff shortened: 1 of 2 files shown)


cmd/app/main.go
```diff
@@ -76,6 +77,17 @@ func main() {
 		log.Panicf("failed to register gateway endpoint: %v", err)
 	}
 
+	if err := d.MUX.HandlePath(http.MethodGet, "/", func(w http.ResponseWriter, r *http.Request, _ map[string]string) {
```
mattmoor, line 80: Are we sure that this won't hide the other handlers?

cpanato: no, i will try to test that,
i was unsure if that is the correct way to do that

mattmoor (approved):

mattmoor (approved):

PR #811, opened 2025-03-21
pin reviewdog/action-tflint github action to full-length commit SHA

eslerm:
Ref: https://github.com/chainguard-dev/prodsec/issues/60

### DIFF
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -83,7 +83,7 @@ jobs:
       - uses: chainguard-dev/actions/eof-newline@main
         if: ${{ always() }}
 
-      - uses: reviewdog/action-tflint@master
+      - uses: reviewdog/action-tflint@92ecd5bdf3d31ada4ac26a702666986f67385fda # master
         if: ${{ always() }}
         with:
           github_token: ${{ secrets.github_token }}


.github/workflows/style.yaml
```diff
@@ -83,7 +83,7 @@ jobs:
       - uses: chainguard-dev/actions/eof-newline@main
         if: ${{ always() }}
 
-      - uses: reviewdog/action-tflint@master
+      - uses: reviewdog/action-tflint@92ecd5bdf3d31ada4ac26a702666986f67385fda # master
```
cpanato, line 86: we should pin that to a spefic tag main is very vague

eslerm: Next cycle ProdSec is writing a SDLC to define GitHub Actions requirements (like pinning). Should we enforce pinning to tags some how? I am not sure if there is a security aspect, but it looks like a good maintenance practice (with some edgecases).

eslerm: I would like the PRs to add full-length commit SHA pins to land before action versions are updated. Action version updates can be done by the maintainers of a repo. These PRs are a security measure to pin existing mutable versions to their equivalent full-length commit SHAs, so that we can avoid future tj-actions style attack.

https://github.com/chainguard-dev/exitdir/pull/25#issuecomment-2743665975

PR #842, opened 2025-04-16
Add support for static JWKS

### LINKED ISSUE #841: Feature request: Ability to trust OIDC tokens bypassing the OIDC protocol with a JWKS file in the GitHub repository
Hi again 👋 

A very cool feature of GCP's Workload Identity Federation is the ability to upload a JWKS file for supporting issuers that are not accessible from the Internet for whatever reasons (compliance, etc.), like a Kubernetes cluster with a private API server (e.g. `kind` running in CI or locally, or a GKE cluster). See the GCP docs for this feature:

https://cloud.google.com/iam/docs/workload-identity-federation#oidc-credential-security

It would be very cool if the trust policy had a field `jwks` where you can specify a multi-line string containing a JWKS JSON document, i.e. the output of a request to `https://<issuer>/openid/v1/jwks`, like this:

```yaml
issuer: https://kubernetes.default.svc.cluster.local # just informative in this case, should be ignored
subject: system:serviceaccount:my-app:my-app
jwks: |
  {
    "keys": [
      {
        "use": "sig",
        "kty": "RSA",
        "kid": "LHVGP8kqzN1MuKRMTsroIcR-7hdicXWdpaquEWcAh9Q",
        "alg": "RS256",
        "n": "s5XuFpodwhj6my_gTUHDKbHmQIx-3Tf40OduMZRWlU6_B_nSdjX01kS1UQSGw_G5eVQARooI-tY1vj3bBwn4dEEFa2TlnNnAJca0hj2Izef8A8Uw-mT0fgGI4Hs3xS84Mn_WXNlKXEiPLiFyOGNr0GQBKZDyTps8JUlvnwuWCv1gkzudUHa8B0i8ITSEUclK9_LqZj4zXUAN0Wj_4DVfI_PQ0IHci9K5Q9bgCV0j1EvTsyrwGyLFwyhktUmNhjREAfgYmxvbIRhPSP4YuO2Et1KM7YmjA75cQ9oE3i-QLrOZDripyMRop5RmWttQCEdEWLQWPzBd7aZ5CLbmZuIlIQ",
        "e": "AQAB"
      }
    ]
  }

permissions:
  contents: read
```

The trust policy above should allow the Kubernetes `ServiceAccount` `my-app` from the namespace `my-app` to read the contents of the GitHub repository.

It would be easy to implement this through `oidc.NewVerifier()`, like you are already doing for tests. This would require moving the call to `s.lookupInstallAndTrustPolicy()` up, above the OIDC token verification part in the `func (s *sts) Exchange()` method.

I'm filing a PR for this 😁

matheuscscp:
Closes #841

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,33 @@ permissions:
   contents: read
 ```
 
+It's also possible to set a static JWKS JSON document for verifying the token
+signature. This is useful for cases where the OIDC discovery endpoint is not
+reachable from the Internet, e.g. a Kubernetes cluster with a private API
+server/OIDC discovery endpoint. Example:
+
+```yaml
+issuer: https://kubernetes.default.svc.cluster.local
+audience: https://kubernetes.default.svc.cluster.local
+subject: system:serviceaccount:my-app:my-app
+jwks: |
+  {
+    "keys": [
+      {
+        "use": "sig",
+        "kty": "RSA",
+        "kid": "LHVGP8kqzN1MuKRMTsroIcR-7hdicXWdpaquEWcAh9Q",
+        "alg": "RS256",
+        "n": "s5XuFpodwhj6my_gTUHDKbHmQIx-3Tf40OduMZRWlU6_B_nSdjX01kS1UQSGw_G5eVQARooI-tY1vj3bBwn4dEEFa2TlnNnAJca0hj2Izef8A8Uw-mT0fgGI4Hs3xS84Mn_WXNlKXEiPLiFyOGNr0GQBKZDyTps8JUlvnwuWCv1gkzudUHa8B0i8ITSEUclK9_LqZj4zXUAN0Wj_4DVfI_PQ0IHci9K5Q9bgCV0j1EvTsyrwGyLFwyhktUmNhjREAfgYmxvbIRhPSP4YuO2Et1KM7YmjA75cQ9oE3i-QLrOZDripyMRop5RmWttQCEdEWLQWPzBd7aZ5CLbmZuIlIQ",
+        "e": "AQAB"
+      }
+    ]
+  }
+
+permissions:
+  contents: read
+```
+
 This policy will allow OIDC tokens from Google accounts of folks with a
 Chainguard email address to federate and read the repo contents.
 
--- a/pkg/jwks/jwks.go
+++ b/pkg/jwks/jwks.go
@@ -1,0 +1,41 @@
+// Copyright 2025 Chainguard, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package jwks
+
+import (
+	"crypto"
+	"encoding/json"
+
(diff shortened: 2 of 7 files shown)


jmeridth:
:robot: (high) AI-assisted review -- 4 models (Opus 4.8, Sonnet 5, Haiku 4.5, Fable 5), verified against `go-oidc v3.20.0` / `go-jose v4`.

First off, apologies this sat without a review for so long, @matheuscscp -- that's on us, not you. Thank you for the contribution and for your patience. The JWKS approach itself is solid: we verified there's no issuer bypass (`CheckToken` enforces iss/sub/aud fail-closed for both paths), no algorithm-confusion or `none` exposure (go-oidc defaults the parse to RS256-only before touching key material), and expiry is enforced in production. Nice tests, too.

This needs a rebase before we review for merge -- and it's a design rebase, not a mechanical one, because of one security issue:

**Order of operations (high).** The change moves `lookupInstallAndTrustPolicy` ahead of token verification (it has to, since the JWKS lives in the policy). That lets an unauthenticated caller send `Bearer <garbage>` + any `scope`/`identity` and drive an installation lookup, a `contents:read` token mint, a GitHub Contents API read, and a revoke -- before any credential is checked. Varying `identity`/`scope` bypasses the trust-policy cache (GitHub rate-limit amplification), and the differing `NotFound` vs `Unauthenticated` responses leak whether a policy exists for arbitrary org/repo/identity. Current main verifies first and then looks up with the verified `tok.Subject`.

The rebase collision is the same problem wearing a different hat: main's `lookupInstallAndTrustPolicy(ctx, scope, identity, subject)` now needs the verified `tok.Subject` for sticky check-run routing (verify -> lookup), while this feature needs the policy before verify (lookup -> verify). The way through is to **split** the lookup: fetch/parse the policy YAML (no subject needed) to pick the verifier, verify the token, *then* do the installation/sticky-routing selection with `tok.Subject`. That keeps all GitHub API work behind a valid signature and resolves the ordering -- one change fixes both the security finding and the conflict.

While you're in there, a few lower-severity items worth folding in:
- Require an exact `issuer` (not just `issuer_pattern`) when `jwks` is set -- `StaticKeySet` tries every key with no `kid` binding, so a loose pattern + multi-key JWKS allows cross-issuer key confusion.
- Set `SupportedSigningAlgs` explicitly in `NewVerifier` (today RS256-only safety rests on an implicit go-oidc default; it also means EC/Ed25519 keys silently fail -- deriving the allowlist from the JWKS keys fixes both).
- Keep `jwksConfigOpts` test-only; if it ever reaches a real constructor param it could disable expiry in prod.

Also rebase-mechanical: the `sts` struct is now `rrm`/`sticky`/`appCount` (no `atr`), and there's an `oidcvalidate.IsValidIssuer` pre-check on the discovery path to preserve. Happy to re-review once it's rebased.

Issue #438, opened 2024-08-07
Add AWS support


mkushakov:
The only way right now to host app is to use Google Cloud Platform as not only infra part relay on GCP specific services, but also some code is really GCP specific, like KMS https://github.com/octo-sts/app/blob/main/cmd/app/main.go#L15

It would be nice to add support for other popular cloud provider like AWS or make it more cloud agnostic.

tuananh:
I plan to fork the repo and do this but it would be nice to refactor this to add support for aws as well.

schlomo:
Any news on this? Considering the same and would be happy to collaborate

ChrisJBurns:
I'd love this also!

wwsean08:
I've opened a PR thru work to implement this, at the very least it implements aws's kms service

sgtoj:
I have self-hosted this on AWS and ran it in "production" for the past several weeks. Initially, I ran it in a container with an ALB in front. Later, I replaced that setup with one running on AWS Lambda using API Gateway v2.

GCP KMS is not required; therefore, this project works on AWS out of the box. You can deploy the [`cmd/webhook`](https://github.com/octo-sts/app/tree/main/cmd/webhook)￼ and the [`cmd/app`](https://github.com/octo-sts/app/tree/main/cmd/app)￼ (i.e., the actual STS). Simply pass the GitHub private key (a.k.a. the app certificate) via the `APP_SECRET_CERTIFICATE_ENV_VAR` environment variable -- this alone removes the GCP KMS requirement. Other components, such as the `cmd/probe`, are not required.

If you want to deploy this on AWS using a serverless architecture, you’ll need handlers that this project does not currently include. To address that, I created the [octo-sts-distros](https://github.com/cruxstack/octo-sts-distros)￼ project. It includes Terraform module that deploys the application using API Gateway v2 and AWS Lambda. As a bonus, it provides an optional, simple setup wizard to quickly register the GitHub App and install it in your organization. The wizard automatically saves the configuration as encrypted AWS SSM parameters. The project even includes a local docker/docker-compose distro if you wish to quick test it out (with the wizard).

sgtoj:
> I'll check out your serverless examples, although I'm not sure using private keys and storing them as environment variables is the way to go given the nature of the private key.

[...]

PR #892, opened 2025-05-23
feat: add support for AWS KMS

wwsean08:
This change implements #438 by creating a new package for interfacing with AWS's KMS service by changing how the KMS service is setup and passed around.  For legacy purposes I added a new environment variable to set the KMS type but have defaulted it to GCP as to not break existing setups that pull in this change.

Testing done:
I've updated and created some unit tests, I would have created more for the AWS KMS service but it doesn't have an interface in the AWS sdk preventing me from mocking it, if you would like me to create tests by messing with the http client  can do that as well.

In addition to that I have set this up and run it against AWS KMS, and have verified it works thru manual testing in a personal test organization using my own octosts app.


cmd/webhook/main.go
```diff
@@ -45,23 +45,23 @@ func main() {
...
-	atr, err := ghtransport.New(ctx, baseCfg, client)
+	atr, err := ghtransport.New(ctx, baseCfg, kmsClient)
 	if err != nil {
 		log.Panicf("error creating GitHub App transport: %v", err)
 	}
 
 	// Fetch webhook secrets from secret manager
 	// or allow webhook secret to be defined by env var.
-	// Not everyone is using Google KMS, so we need to support other methods
+	// Not everyone is using Google kmsProvider, so we need to support other methods
```
mkushakov, line 64: adding `github.com/octo-sts/app/pkg/kms` is actually an attempt to add support for more KMS providers, isn't it?

pkg/awskms/awskms.go
```diff
@@ -0,0 +1,69 @@
+// Copyright 2024 Chainguard, Inc.
```
cpanato, line 1:
```suggestion
// Copyright 2025 Chainguard, Inc.
```

pkg/awskms/awskms_test.go
```diff
@@ -0,0 +1,28 @@
+// Copyright 2024 Chainguard, Inc.
```
cpanato, line 1:
```suggestion
// Copyright 2025 Chainguard, Inc.
```

pkg/kms/kms.go
```diff
@@ -0,0 +1,73 @@
+// Copyright 2024 Chainguard, Inc.
```
cpanato, line 1:
```suggestion
// Copyright 2025 Chainguard, Inc.
```

pkg/kms/kms_test.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2024 Chainguard, Inc.
```
cpanato, line 1:
```suggestion
// Copyright 2025 Chainguard, Inc.
```

cpanato (changes requested):
looks nice, I just would like to organize the code better then as you are creating a new kms provider

I was thinking in 

```
- pkg
   |
   - kms
     | 
     - aws
     - gcp
     kms.go
     kms_test.go
```

something similar we do in https://github.com/sigstore/sigstore/tree/main/pkg/signature/kms

thanks for working on this

cmd/webhook/main.go
wwsean08: thanks, you're right, that got caught by a refactor and I missed it

wwsean08:
I've pushed up the requested changes, I will push up a separate PR to update the boilerplates in the hack directory to change the year as to not mix changes together

wwsean08:
Hey @ChrisJBurns , just letting you know, I haven't forgotten, I realized when trying to set it up in AWS that since I never bothered to setup the webhook I need to make additional changes to make that possible, which I will be working on this week.

ChrisJBurns:
@wwsean08  Cheers bud! I'm getting more and more time to look at this on our-side too and what it would look like. Whether its a Kubernetes variant (most likely for now) or possible an ECS variant. Happy to hear that there are others also looking at the AWS side of things! 🚀

wwsean08:
Ok I finished writing up some quick refactoring to handle both AWS and GCP, I need to get them reviewed internally and test them before publishing the changes, but expect them either later this week or early next week, I'm currently on call and handling the support queue, so it's a busy week :)

ChrisJBurns:
@wwsean08  So i've been making some progress on the AWS ECS front. I've managed to get the main `app` container running. I've given it dummy env vars for now, but _slowly_ getting there. The issue I've had so far is that I realise the images themselves on `cgr.dev` and aren't free to use. For your changes, I had to create a standard `Dockerfile` and just do a go build. But realistically it seems that i'll have to have custom image builds as the `latest` versions on `cgr.dev` aren't available.

wwsean08:
So I've just pushed up a change that has a working* webhook, which you can see below.

Screenshot 2025-06-07 at 10 27 07 PM

I have that asterisks because I somehow with two changes hit API rate limits due to it trying to refresh the installation id token, and I'm unsure of why it's spamming that endpoint, but it doesn't _seem_ like something that I introduced, but will keep digging in.   I also looked over the probers and am fairly confidence the way they are coded won't work for anyone outside of chainguard to host the images, and if you wanted them you'd have to code up your own to be slightly different, pointing at your own repo/policies since the repo that it probes against is internal/private so we can't really see the contents of those policies at a minimum, and your github app likely won't be installed into chainguard's org thus not actually making it a valid test, though I'll defer to the maintainers on my analysis of that.

Chainguard does have some free images, I used them to build my test setup in EKS, here's an example of my dockerfile for building the app

```dockerfile
FROM cgr.dev/chainguard/go AS builder
COPY . /app
ENV CGO_ENABLED=0
RUN cd /app/cmd/app && go build -ldflags '-s -w' -o app .

FROM cgr.dev/chainguard/static
COPY --from=builder /app/cmd/app/app /usr/bin/
ENTRYPOINT ["/usr/bin/app"]
```

ChrisJBurns:
@wwsean08  The rate limiting stuff is interesting!! I'm also wondering if the eventing functionality is needed for AWS too? As I think that's only relevant for GCP but it still forces the user to provide a variable to it.

mattmoor:
You don't need a Dockerfile for this, see here:
https://github.com/octo-sts/app/blob/e463b50f88af31c448ef262d488650b76ed6ca50/iac/main.tf#L21-L24

If you are using K8s manifests than I'd just use `ko` directly vs. via TF.

ChrisJBurns:
@wwsean08 That's interesting because as far as I can see, you've not modified any event code.

ChrisJBurns:
I've given this a go on ECS and I'm getting the following in the main `app` container:

```
INFO exchange request: &v1.ExchangeRequest{state:impl.MessageState{NoUnkeyedLiterals:pragma.NoUnkeyedLiterals{}, DoNotCompare:pragma.DoNotCompare{}, DoNotCopy:pragma.DoNotCopy{}, atomicMessageInfo:(*impl.MessageInfo)(*****)}, sizeCache:0, unknownFields:[]uint8(nil), Aud:[]string(nil), Scope:"$ORG/$REPO", Identity:"foo", Cap:[]string(nil), IdentityProvider:""}
```

But nothing tends to happen after that. I get `Attempt 1 failed. Error: HTTP error! status: 500` logs in the Github Action too even though I get a `200` when I go to both the domain and webhook endpoint. The webhook also has tonnes of `2025/06/19 22:06:29 ERROR error validating payload: mime: no media type X-GitHub-Delivery="" X-GitHub-Event=""` errors but I think that's just the health check - although at this point I'm not too sure.

ChrisJBurns:
@wwsean08  I can confirm that I _think_ I'm getting the same issue as you. My "Trusty Policy" passes but my pipeline action fails due to a `500` - still not sure why I'm getting `500` but I'm more interested in the looping over the checkruns that the webhook is doing.

I'm seeing a bunch of `INFO created CheckRun X-GitHub-Delivery=........` in the webhook logs.

Here is my actions associated with a PR
image

mattmoor:
I think this is a @wlynch question

ChrisJBurns:
@wlynch just chasing the above incase it got lost in the notification craziness

pkg/kms/kms.go
```diff
@@ -0,0 +1,73 @@
...
+	GCP = "gcp"
+)
+
+type kmsProvider struct {
+	ctx       context.Context
+	provider  string
+	kmsKey    string
+	gcpClient *kmsGCP.KeyManagementClient
+	awsClient *kmsAWS.Client
+}
```
wlynch, line 30: Instead of storing all of the clients here, have each impl have a function that returns a struct that implements the KMS interface. Then we can do something like:

```
func NewKMS(ctx context.Context, provider, kmsKey string) (KMS, error) {
	switch strings.ToLower(provider) {
	case GCP:
		return gcp.NewProvider(ctx, kmsKey)
	case AWS:
		return aws.NewProvider(ctx, kmsKey)
	default:
		return nil, errors.New("unsupported kms provider")
	}
}
```

Alternatively, skip the KMS interface altogether and create funcs along the lines of `gcp.NewDefault(ctx, key)` that do the same thing here with default clients initialized.

wlynch:
https://github.com/octo-sts/app/pull/988 for filtering out what I believe are the problematic webhook events

ChrisJBurns:
@wwsean08 @wlynch  Amazing thanks folks! I've been away this week but I plan on giving this another try on the weekend with only push and PR events!

ChrisJBurns:
@wwsean08 I've got some changes made to implement the above feedback by @wlynch , not sure the best way of submitting them. I could either raise a PR against this repo (would mean a duplicate PR though), or raise it against your fork? To do the latter, I'd need to have permissions to push to a branch 😄

ChrisJBurns:
@wlynch I believe @wwsean08 has merged my PR into his branch now, if you could give it another pass that would be amazing!

ChrisJBurns:
Bumping this incase its got buried in notifications

PR #901, opened 2025-05-28
feat: Add organization-wide trusted token issuers functionality

### LINKED ISSUE #900: Feature Request: Organization-wide Trusted Token Issuer Validation
## Problem Statement

Currently, octo-sts validates OIDC tokens against individual trust policies stored in `.github/chainguard/{name}.sts.yaml` files. While this provides repository-level security, organizations lack a centralized mechanism to control which OIDC issuers are permitted to federate with their repositories at the organization level.

This creates potential security gaps where:
- **Malicious or compromised OIDC providers** could potentially be used if they satisfy individual trust policies
- **Organizations cannot enforce issuer standards** across all repositories
- **No defense-in-depth** for OIDC issuer validation
- **Difficult to audit** which issuers are trusted across the organization

## Proposed Solution

Add support for **organization-wide trusted token issuer validation** as an additional security layer that validates OIDC issuers before individual trust policy evaluation.

### Implementation Overview

1. **Configuration Location**: `.github/chainguard/trusted-token-issuers.yaml` in the organization's `.github` repository
2. **Validation Flow**: Organization issuer validation → Individual trust policy validation
3. **Backward Compatibility**: When disabled or missing, all issuers are accepted (current behavior)

### Configuration Format

```yaml
# Organization OIDC Trusted Token Issuers Configuration
description: "Organization OIDC trusted token issuers configuration"
enabled: true

# List of exact trusted issuer URLs
trusted_issuers:
  - "https://token.actions.githubusercontent.com"
  - "https://accounts.google.com"

# List of issuer regex patterns for flexible matching
issuer_patterns:
  # AWS EKS OIDC issuers
  - "https://oidc\\.eks\\.[a-z0-9-]+\\.amazonaws\\.com/id/[A-Z0-9]+"
```

### Security
- **Defense-in-depth**: Adds an additional validation layer before trust policy evaluation
- **Organization-wide control**: Centralized management of permitted OIDC issuers
- **Prevents unauthorized issuers**: Blocks federation attempts from unexpected or compromised OIDC providers

### Operational
- **Git-based audit trail**: Configuration changes tracked through Git history and pull requests
- **Flexible matching**: Support for both exact matches and regex patterns
- **Zero breaking changes**: Fully backward compatible with existing deployments

### Compliance
- **Centralized governance**: Helps organizations meet security compliance requirements
- **Auditable**: Clear trail of what issuers are trusted and when changes were made

### Core Components
- `OrgTrustedTokenIssuersValidator` - Main validation logic with LRU caching
- `OrgTrustedTokenIssuersConfig` - Configuration structure with YAML parsing
- URL format validation for security (HTTPS enforcement, malformed URL detection)
- Comprehensive test suite (38+ test cases covering all scenarios)

### Performance Features
- **LRU Cache**: 50 entries with 5-minute TTL to minimize GitHub API calls
- **Graceful degradation**: Continues to work if GitHub API is unavailable
- **Efficient regex compilation**: Patterns compiled once and cached

### Security Features
- **HTTPS enforcement**: Only HTTPS issuers are accepted
- **URL validation**: Rejects malformed URLs and suspicious patterns
- **Pattern safety**: Invalid regex patterns cause configuration rejection

## Backward Compatibility

✅ **Zero breaking changes**
- When `enabled: false` or configuration missing, all issuers are accepted
- Existing trust policies continue to work unchanged
- No changes to existing API or behavior

## Implementation Status

- ✅ Core implementation complete (`pkg/octosts/org_trusted_token_issuers.go`)
- ✅ Comprehensive test suite (100% pass rate)
- ✅ Documentation updated (README.md)
- ✅ Example configuration provided
- ✅ Backward compatibility verified

### GitHub Actions Only Organization
```yaml
enabled: true
trusted_issuers:
  - "https://token.actions.githubusercontent.com"
```

### Multi-Cloud Enterprise
```yaml
enabled: true
trusted_issuers:
  - "https://token.actions.githubusercontent.com"
  - "https://accounts.google.com"
issuer_patterns:
  - "https://oidc\\.eks\\.[a-z0-9-]+\\.amazonaws\\.com/id/[A-Z0-9]+"
  - "https://login\\.microsoftonline\\.com/[a-f0-9-]+/v2\\.0"
```

## Testing

The implementation includes comprehensive testing:
- URL format validation (valid/invalid/malformed issuers)
- Pattern compilation and matching
- Cache management (LRU eviction, TTL expiration)
- Configuration loading and error handling
- Real-world pattern matching (AWS EKS, Azure AD, GCP)

## Migration Path

1. **Deploy**: Feature ships disabled by default
2. **Configure**: Organizations create configuration with `enabled: false` 
3. **Test**: Validate configuration works in development
4. **Enable**: Set `enabled: true` and monitor
5. **Refine**: Adjust patterns based on usage

## Alternatives Considered

1. **Environment variables**: Less flexible, harder to audit
2. **Database storage**: Adds infrastructure complexity
3. **API-based configuration**: Requires additional authentication mechanisms

Git-based configuration was chosen for auditability, pull request workflows, and consistency with existing trust policy management.

## Questions for Maintainers

1. Are there any concerns with the proposed configuration file location?
2. Should we consider making the cache size/TTL configurable?
3. Any preferences for the configuration schema or field naming?

---

**PR submitted**: This feature is fully implemented with comprehensive tests and documentation. PR at https://github.com/octo-sts/app/pull/901.

**Files involved**:
- `pkg/octosts/org_trusted_token_issuers.go` (implementation)
- `pkg/octosts/org_trusted_token_issuers_test.go` (tests)
- `README.md` (documentation)

### LINKED ISSUE #1546: Org allowlist: enforcement gap closed by #901; residual limitations recorded
## Summary

#1543 adds `ghinstall.Manager.GetAllFresh` and routes `manager`'s cache writes through `setInstalled` / `setNotInstalled`. Both stand on their own — the cache fix repairs a routing bug reachable on `main` today — but **`GetAllFresh` ships with no production caller.**

The caller is the reason the method exists, and it is deliberately not in #1543 or #901. This issue tracks that gap so the exported API does not sit unused indefinitely, and so the fail-open it was written to close does not get forgotten.

## What is missing

**1. No production consumer of `GetAllFresh`.** After #1543 merges, the only callers are tests. It is on the `Manager` interface, so every implementation (including five test fakes) carries a method nothing in the shipping code path invokes. That is defensible for one release — #1525 shipped `GetAll` the same way — but it should not be the steady state.

**2. `confirmNoAccess`, the consumer.** This is the connecting tissue. Without it, `orgIssuerLookup` can cache org-wide allow-all off an enumeration that was silently incomplete:

- An organization runs App A on selected repositories only, so A cannot read `.github`.
- It installs App B specifically to turn issuer enforcement **on**.
- `manager.GetAll` maps a `NotFound` to `(nil, nil)`, and a `NotFound` served from the five-minute negative cache is indistinguishable from a real one — so `GetAll` returns `([A], nil)`: non-empty, nil error, silently missing B.
- Every installation it can see reports no access, so `Absent` — permit all issuers — is cached for the whole organization, at exactly the moment enforcement was meant to begin.

The existing "at least one installation answered" guard catches an *empty* enumeration. It does not catch a non-empty but incomplete one. And it is reachable rather than theoretical: `roundRobin.Get` walks the other Apps on `NotFound`, so ordinary exchange traffic keeps App B's negative entry warm right up until B is installed. The stale-entry fallback does not help either, because a first-time enforcement setup has no prior good entry to fall back to.

`confirmNoAccess` calls `GetAllFresh` before concluding absence and reads any installation the first pass could not see. If one can read `.github`, enforcement turns on in that same request; if they all agree, the conclusion is earned. A failed confirmation is ignorance, not absence, and falls through to stale / fail-closed.

## Why it was split out

To make #1543 and #901 independently mergeable. `confirmNoAccess` is the only thing in either PR that couples them — a single call to `s.rrm.GetAllFresh`. Keeping it in #901 would have made #901 unmergeable until #1543 landed. So both PRs now stand alone, and the tissue waits.

The cost, stated plainly: **until this lands, an organization adopting the trusted-issuer allowlist carries a window of up to the negative-cache TTL (5 minutes) after installing an App, during which enforcement can be silently off.** It is bounded, it self-heals, and it is strictly better than the status quo of no allowlist at all — but it is a real gap in a security control, and that is why this issue exists rather than a TODO comment.

## Definition of done

- [ ] `GetAllFresh` has a production caller
- [ ] `orgIssuerLookup` confirms before caching `Absent`, and reads any newly-visible installation rather than treating a discrepancy as unknown
- [ ] A failed confirmation caches nothing
- [ ] A successful confirmation is bounded by the `orgIssuers` TTL — at most one per owner per TTL, not one per exchange
- [ ] Docs state what confirmation does *not* cover (see below), so no one reads "confirmed against GitHub" as "the window is closed"

The work exists on `feat/octosts-confirm-no-access` and is blocked only on #1543 and #901 merging; it will be opened as a PR at that point, when its diff is genuinely just the ~460 lines of tissue rather than the sum of its two dependencies.

## Known gaps that remain even after this closes

Worth recording here so they are not mistaken for oversights later:

**GitHub's replication, not just our cache.** `GetAllFresh` bypasses *this service's* negative cache. It cannot make `ListInstallations` return an installation GitHub has not yet propagated, so a confirmation can still agree with an incomplete first pass. That narrows the window rather than eliminating it. Closing it entirely would mean failing closed for every single-App organization that legitimately cannot read `.github` — the common case — so it is accepted and documented.

**The cache invariant is narrowed, not established.** `setInstalled` orders its two writes so a concurrent reader re-walks rather than reading a stale negative entry, but `manager` has no mutex and check/walk/commit is not atomic. A `Get` whose walk finished before an install can still commit "not installed" afterwards and re-shadow the entry. Closing that needs per-owner serialization or versioned observations. It fails safe — the result is a spurious `NotFound`, so routing rejects exchanges rather than granting anything.

**The positive cache has no TTL.** An uninstalled App keeps appearing in enumerations. That direction fails closed (its token mint fails, which counts as a failure rather than an agreement), so it does not feed the fail-open. Making enumeration authoritative means invalidating on installation webhooks, which is a separate change.

**Stale `Absent` can outlive its documented hour.** Serving a stale entry re-inserts it into the five-minute cache, so an `Absent` that entered both caches at `t=0` can still be served at roughly `t=64:59`. Relevant here because the stale path runs *before* any confirmation, so an organization previously confirmed-`Absent` that then installs an App can sit on served-stale allow-all through a `GetAllFresh` outage. Capping the reseed at the stale entry's original expiry, or serving stale without reseeding, would fix it.

salekseev:
Closes #900. Closes #1546.

Three residual limitations in installation enumeration move to #1550. Each one fails safe today, and #1550 records which assumption each one rests on.

This replaces the original implementation in this PR entirely. The previous approach could not work as written, and the reasons shaped the design that follows — see [Why this is a rewrite](#why-this-is-a-rewrite).

## What it does

An organization restricts which OIDC issuers may federate with any of its repositories by committing `.github/chainguard/trusted-token-issuers.yaml` to its `.github` repository:

```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/octo-sts/app/refs/heads/main/pkg/octosts/octosts.OrgTrustedIssuers.json

mode: audit                 # enforce (default) | audit

issuers:
  - https://token.actions.githubusercontent.com
  - https://tenant.us.auth0.com/          # a trailing slash is a distinct issuer

issuer_patterns:
  - https://oidc\.eks\.[a-z0-9-]+\.amazonaws\.com/id/[A-Z0-9]+
```

When the file is present, a token whose issuer is not permitted is rejected **before its trust policy is read**. `issuers` matches exactly; `issuer_patterns` are anchored regular expressions that must match the whole URL; an issuer passes if it matches either list.

**No file means no restriction.** Behavior is unchanged for every organization that does not add one.

The problem this addresses: today anyone who can write a trust policy in any repository can point `issuer:` at an identity provider they control. The org owner has no way to constrain that centrally.

## Why this is a rewrite

The original implementation here had two defects that could not be patched around:

1. **It read the config with the App JWT.** An App JWT authenticates the App, not an installation, and cannot read repository contents. The lookup could never have succeeded against real GitHub.
2. **It failed open on every error.** Any GitHub error — including a rate limit — silently disabled enforcement. A control that an attacker can switch off by inducing an error is not a control.

Fixing (1) means minting an installation token on the token-exchange hot path, which forces an explicit decision at every failure mode rather than one blanket fallback. That is most of what this PR is.

## Design

**A dedicated file, presence as the switch.** No new field on `OrgTrustPolicy`, no operator configuration to enable it. Committing the file turns it on; deleting it turns it off. This keeps the control legible to an org owner reading their own repository, at the cost that deleting the file fails open (see Hardening in the README).

**`audit` mode for rollout.** In audit mode nothing is rejected; each token that *would* have been rejected is logged and recorded on the exchange event as `issuer_allowlist`, so an org can confirm the list is complete before enforcing. This is the only safe way to adopt the feature on a busy org.

**One parse path for both consumers.** `octosts.ParseOrgTrustedIssuers` unmarshals strictly *and* compiles. Both the exchange path and the webhook check run call it, so a green check run cannot disagree with what happens at exchange time. `pkg/webhook` also uses the exported `octosts.OrgTrustedIssuersPath` rather than its own copy of the path.

**Least-privilege read.** The file is read with a short-lived token scoped to exactly `{repositories: [".github"], permissions: {contents: read}}`. That scope is pinned by a test which decodes the minted token request off the wire, so widening it fails CI.

**GitHub Enterprise Server.** The allowlist read honors the configured API base URL from #1400. Getting this wrong is not a cosmetic bug: a read that goes to `api.github.com` on a GHES deployment fails, is classified as "this installation cannot see `.github`", and therefore permits every issuer — a fail-open across the entire deployment. Rebasing onto #1400 is what surfaced it, since nothing in the test suite could: the fakes inject a transport that rewrites requests to a local server, so the base URL is never consulted. The four duplicated client-construction blocks that allowed a fourth copy to drift are collapsed into one `newGitHubClient` helper, with a test enforcing it is the package's only constructor — split out and merged as #1529, since it is a refactor of pre-existing duplication.

**Patterns are anchored *and* boundary-checked.** Anchoring the ends is not sufficient — a pattern loose in the middle bypasses the control regardless. Every single-character atom in an `issuer_patterns` entry is therefore tested against `/`, and the pattern is rejected if any atom can match one. That is decided empirically (compile the atom, test it) rather than by banning named constructs, because an explicit range such as `[.-9]` contains `/` at 0x2F and no blacklist of names would catch it. Literal `/` is fine, so a path is written one segment at a time. **This is a best-effort guard against accidents, not a security boundary**

**Four bypasses in this guard were found and fixed by an adversarial review pass, and the guard now fails closed.** Each of these passed validation while matching `/`:

```
https://[^]]+\.example\.com      permitted https://evil.com/x.example.com
https://a\x2Fb\.example\.com     permitted https://a/b.example.com
https://a\x{2F}b\.example\.com   permitted https://a/b.example.com
https://a\057b\.example\.com     permitted https://a/b.example.com
```

Two parsing gaps: RE2 reads a `]` in the first position of a class as a literal member, so `[^]]` is "any byte except `]`" — but the scanner stopped at that first `]` and extracted the uncompilable `[^]`. And `\x2F` / `\x{2F}` / `\057` *spell* a character rather than naming a class, so they fell to the escaped-literal branch, which skipped two bytes and left the digits as unexamined literal text.

The root cause mattered more than either gap: on a failed compile of the extracted atom, the scanner did `continue`. A validator that silently skips what it cannot parse is fail-**open**, so every present and future parsing gap was a bypass by the same route. An unanalyzable atom now rejects the pattern, and `TestCompileFailsClosedOnAnAtomItCannotAnalyze` pins that posture rather than any one syntax. All four cases are rows in the existing bypass table, which asserts each pattern really does match its demonstration issuer so no row can go vacuous. Verified against false rejections: literal-`]` classes, POSIX classes, and the real EKS and Keycloak patterns from the fixtures and README all still pass.

Worth being straight about the severity: since the file is written by the org owner, this was not attacker-controlled input, so it was a footgun rather than a remote bypass. But the docs told an owner that every atom is tested against `/` and the pattern rejected if any can match one, and that was not true — so a validated pattern carried trust it had not earned. — the file is written by an org owner, the principal the control serves, so an over-broad pattern is an unwise choice rather than an attack. See [Adversarial review](#adversarial-review).

### Failure modes

Every one of these is a deliberate choice, and the fail-open ones are the load-bearing part of the review:

| Situation | Result |
|---|---|
| File absent (definitive 404) | All issuers permitted |
| No installation can read `.github` | All issuers permitted, logged as a warning |
| Rate limited / GitHub unavailable | Last known good allowlist; if none, exchange rejected (`ResourceExhausted`) |
| File present but invalid | Last known good allowlist; if none, every exchange in the org rejected (`FailedPrecondition`) |
| Enumeration incomplete or partly failed | Last known good; if none, exchange rejected (`Unavailable`) |

The distinction that makes this sound: **`Absent` requires positive knowledge.** It is cached only from a definitive 404, or from an exhaustive enumeration in which *every* installation reported it cannot read `.github`. "Every installation agrees it cannot see the file" is a known state and falls open. "We could not establish that" is unknown and falls closed. A partial or failed enumeration is never cached.

Two timing consequences worth knowing before adopting: after *narrowing* an allowlist, the previous broader list can still be served for up to an hour whenever GitHub errors; and after *fixing* an invalid file, rejections can persist for up to five minutes while the cached result expires.

### Why enumerate installations rather than route to one

`ghinstall.Manager.Get` is a routing primitive and cannot enumerate — see #1527 for the full mechanics. Briefly: with warm quota data `pickFromPool` selects `argmax(remaining)` within a tier, deterministically and without rotation; without it the shared counter rotates but guarantees no coverage under concurrency; and the multi-app fallback fires only on `codes.NotFound`, not on the 422 a `.github`-scoped mint returns when the App lacks access. A retry loop built on it could see one blind installation N times and conclude that no installation can read `.github` — caching allow-all for the whole organization.

`Manager.GetAll`, split out as #1525, provides exhaustive enumeration. Its contract is explicit that a non-nil error means the enumeration is **not** exhaustive even when the returned slice is non-empty, because callers that need exhaustiveness for correctness must treat that as unknown. `TestLookupSurvivesOneBlindInstallation` is the regression test: the first installation cannot read `.github`, a second can and holds an enforcing allowlist, and because `Get` always returns the first, only an enumeration-based lookup finds the second.

### On `GetAll` being exhaustive only over observed state

The review note about this has two parts, and both are **already handled here**.

A nil error with an empty result is not read as absence: `exhaustiveNoAccess` requires at least one installation to have answered, because with none the count comparison degenerates to `0 == 0` and would grant allow-all off no evidence at all. And per-installation read errors already fail closed — only an explicit "cannot see `.github`" counts toward agreement, so any failure or rate-limit leaves the enumeration non-exhaustive and the exchange is rejected.

Working on that, I found a **third** case neither of us named: a non-empty but *incomplete* enumeration. `manager.GetAll` maps a `NotFound` to `(nil, nil)`, and a `NotFound` served from its five-minute negative cache is indistinguishable from a real one — so an org running App A (which cannot read `.github`) that installs App B to turn enforcement **on** gets `([A], nil)` for up to five minutes: non-empty, nil error, silently missing B. Every installation it can see reports no access, so allow-all is cached org-wide at exactly the wrong moment.

**That fix is deliberately not in this PR**, and is tracked by #1546. It needs `GetAllFresh` from #1543, and including it would make this PR unmergeable until #1543 lands. It is a self-contained ~350-line change that I will open as a follow-up once #1543 and this PR are both in, so all three review independently. Until then, adopting this feature carries that five-minute window; it is bounded, it self-heals, and it is strictly better than the status quo of no allowlist at all.

## Split out of this PR

Per @jmeridth's request, the independently-reviewable pieces are now their own PRs. Each is mergeable on its own and green standalone. **Suggested review order: #1524, #1529, #1531, #1525, then this one** — the first four are small and independent, and this PR's diff shrinks to just the feature once they land.

**All four are now merged, and this branch has been rebased onto current `main`.** Their commits are out of this diff.

**A fifth is split out: #1543** — `pkg/ghinstall`'s `GetAllFresh` plus a negative-cache shadowing fix, 476 lines across 4 files. (I had described that fix as "reachable on `main` today"; it is not, sequentially — see #1543 for the correction. It is a prerequisite for `GetAllFresh` rather than a standalone bug fix.) Same reasoning as #1525: an enumeration primitive plus a routing-cache bug, neither of which mentions trusted issuers.

**This PR should merge BEFORE #1543.** I originally wrote "either order" here and that was wrong — I verified it by actually merging them both ways locally, and whichever goes second leaves `main` with a broken test build:

#1543 adds `GetAllFresh` to the `ghinstall.Manager` interface. This PR adds two fakes that implement `Manager` (`enumMgr` and `notInstalledMgr`, in a file that does not exist on `main`, so #1543 cannot stub them). Once both are in, those fakes no longer satisfy the interface:

```
vet: org_trusted_issuers_store_test.go:786: cannot use (*enumMgr)(nil) as ghinstall.Manager
     value in variable declaration: *enumMgr does not implement ghinstall.Manager
     (missing method GetAllFresh)
```

Worth stressing how quietly this fails: `go build ./...` passes either way, there is no textual conflict, and GitHub reports both PRs MERGEABLE. Only the test build breaks, so the first merge is green and the damage appears on the second.

The fix is small and belongs to #1543, not here: after this PR merges, #1543 adds the two stub methods to `org_trusted_issuers_store_test.go` — which by then exists on `main`, and where they are *used* (they satisfy the interface #1543 itself introduces), so they do not trip the `unused` linter. They cannot be added here preemptively for exactly that reason: without #1543's interface method they are unused methods on unexported types, which fails `golangci-lint`.

Reviewing the two independently is still fine. It is only the merge order that is constrained.

The one piece that needed both PRs — the confirmation step described above — is held back for a follow-up rather than coupling them.

This PR is also squashed to **4 commits**, each of which builds and passes its tests on its own, so the change is readable in stages and bisectable.

Three of the rebase conflicts needed a judgement call rather than a mechanical resolution, and are worth knowing about when reading the diff:

- **`isValidatedPath`** — `main`'s version from #1531 is the feature-stripped one; this branch's takes a `repo` argument and also matches the org allowlist path. I kept `main`'s nested-`if` call sites and its richer comments and added only the argument, so the diff shows the feature rather than a restyle.
- **`pkg/octosts/client_test.go`** — resolved as an add/add. `main`'s version is a strict superset, so I took it wholesale.
- **`TestIsValidatedPath`** — this branch's repo-aware version replaces `main`'s, but had dropped two of `main`'s rows (a `README.md` and a non-policy `.yaml` inside `.github/chainguard`). I restored both, pointed at the `.github` repository where *both* arms of the predicate are live. Without them, loosening the allowlist check to a `.github/chainguard/*.yaml`-style glob would pass every test in the suite.

| PR | Issue | What |
|---|---|---|
| #1524 | #1526 | Rate-limit classification fix. A pre-existing bug: a real primary rate limit yields `*github.RateLimitError`, which does not unwrap to `*github.ErrorResponse`, so the stale-cache fallback and cross-app retry were both dead in production. |
| #1525 | #1527 | `ghinstall.Manager.GetAll`. `Get` is a routing primitive and cannot enumerate, so a caller surveying installations can be answered by the same one repeatedly. |
| #1529 | #1528 | One GitHub client constructor in `pkg/octosts`. The duplicated construction block is what let a fourth copy drift and lose the GHES base URL — a defect no test could catch, since the fakes bypass the base URL entirely. |
| #1531 | #1530 | Webhook path-predicate consolidation, plus filtering the unfiltered directory scan. **Contains the behavior change** described below. |

## Behavior change to an existing path

`handleCheckSuite`'s new-branch/initial-commit path listed `.github/chainguard` and validated **every** entry with no filtering. It now filters through the same predicate as every other path. A file in that directory that is not a trust policy — a `README.md`, for instance — was previously fetched and parsed as a `TrustPolicy`; now it is skipped. This is a fix, but it is a behavior change on a pre-existing path and worth flagging explicitly.

The five inline `filepath.Match` sites are also consolidated into one `isValidatedPath(repo, path)` predicate, which is now the only glob in `pkg/webhook`.

## Resolved: `.vscode/settings.json` is untouched

**This PR no longer changes `.vscode/settings.json`** — it is reverted to `main` and out of the diff, per @jmeridth's call.

And a correction to my earlier reasoning here, which was wrong: I claimed the README points at the settings file. It does not. Since #1270 the README recommends the inline `# yaml-language-server: $schema=...` header, and the examples use it directly — so there was never a README dependency on the committed editor config to keep in step.

What the allowlist gets instead is that same inline header: on the README example, and now on all four testdata fixtures pointing at `octosts.OrgTrustedIssuers.json`. While adding those I also gave the fixtures the copyright headers every other testdata YAML in the repo carries and which I had omitted.

Removing `.vscode/settings.json` from the repo entirely — now redundant with the inline-header approach, and editor-specific — is tracked separately by @jmeridth rather than smuggled in here.

One consequence worth stating plainly, since it is a real if minor regression I am choosing not to fix here: `main`'s glob is `/.github/chainguard/*`, which matches everything in that directory. So for anyone still using the committed settings file, an editor will validate `trusted-token-issuers.yaml` against the **TrustPolicy** schema and report errors on a valid file. The inline header in the fixtures and the README example takes precedence in vscode-yaml, so following the documented approach avoids it; the separate removal issue closes it properly.

## Deliberately out of scope

**An operator-level env var** (the deployment-wide equivalent) is left to #1396. This PR is org-owner-controlled configuration only. The reasoning: authorization at the org-owner level already governs who can write `org/.github`, so the org-level file is coherent on its own, and the two features have different threat models and different audiences.

## Testing

Every behavioral claim above has a test, and the tests were validated by mutation testing rather than by coverage: for each security-relevant branch, the implementation was deliberately broken to confirm a test actually fails. That process found several tests that passed under both the correct and the broken implementation, and those were rewritten. Notable examples now pinned:

- a contents 403 is one installation's blindness, not org-wide absence
- a stale `Absent` entry is never substituted for a currently-invalid config (that would convert a broken file into allow-all)
- one blind installation plus one failed request is `Unavailable`, not `Absent`
- the definitive-404 path resolves to `Absent` — the default state of every org that has not adopted the feature
- the minted token scope is exactly `.github` + `contents: read`
- the allowlist verdict is cached, asserted by observed GitHub read count rather than by cache occupancy
- an audit-mode decision reaches the emitted CloudEvent, which is the only operator-visible signal for a pass-through
- an `issuer_patterns` entry cannot use a wildcard that crosses a DNS label or a path separator, asserted against the concrete issuers such a pattern would otherwise permit

`go test ./... -race`, `-shuffle=on`, `gofmt`, `go vet`, and `golangci-lint run ./...` (`0 issues`) all pass.

## Adversarial review

I ran several independent adversarial reviews over the finished branch, each briefed to attack the design rather than hunt for defects. **Every one of them flagged the issuer-pattern matcher as the weakest point, and every one found something there** — the guard has now been revised four times, and each revision changed approach rather than lengthening a blacklist. The remaining objections are design trade-offs worth your judgement; I have not silently overruled them.

### Models used

| Pass | Model | Outcome |
|---|---|---|
| Design review, pre-split | `opus-5` (xhigh) | Flagged the matcher; drove the first two revisions (anchoring, then per-atom boundary checking) |
| Adversarial, full branch | `gpt-5.6-sol` (high) | No defect introduced by the change; two pre-existing findings, both now in the follow-ups list |
| Adversarial, full branch | `kimi-k3` (max) | **REQUEST CHANGES.** Found the `[[:/-~]` bypass, the read-403 misclassification, and the webhook case-fold asymmetry |

The second `kimi-k3-max` run was briefed that the first four bypasses were already fixed and told to look for a fifth that the fail-closed posture would not catch — which is what it found. It also investigated the >1 MB contents-file path and explicitly declined to report it, having confirmed GitHub returns 200 with `encoding: "none"` so the code already fails closed. I mention that because a review that reports its negatives is worth more than one that doesn't.

Six pattern bypasses were found and fixed in total: unanchored matching, top-level alternation, atoms that match `/` by name or by range, the leading-`]` class, numeric escapes spelling `/`, an unbalanced `)` escaping the anchoring group, and `[:` without `:]`. The pattern that emerged is that **per-atom analysis is only sound while the scanner tokenizes the way RE2 does** — four of the six were tokenization divergences, not missing cases, which is why the guard now also fails closed on any atom it cannot compile.

### Fixed twice: operator-written patterns could cross DNS and path boundaries

An operator writing what reads as "any subdomain of example.com":

```yaml
issuer_patterns:
  - https://.*.example.com
```

also permitted `https://evil.attacker-example.com` (`.*` consumes `evil.attacker`, and the unescaped `.` matches the `-`) and `https://totally-evil.com/x?a=example.com` (`.*` crosses the `/` too). Both verified returning `true` before the fix. A repository writer could then point their trust policy at that attacker-controlled issuer and defeat the org control entirely — via the most natural way to express the intent.

`Compile()` now rejects `.` used as a wildcard, with an error naming the pattern index, the two issuers it would wrongly permit, and the corrected form. The more likely real-world shape is also covered: `https://[a-z0-9-]+.example.com`, where an operator has bounded the varying label but not escaped the separator, permits the registrable `https://evil-example.com`.

**The first fix banned `.` and was not enough.** The second review showed four more spellings of "any character" that passed it and matched `https://totally-evil.com/x.example.com` — `\S+`, `[\s\S]+`, `[^\n]+`, `[[:ascii:]]+` — all verified. Worse, my error message told operators to use "an explicit character class", and two of those *are* explicit character classes. The guard advertised protection it did not provide.

So the approach changed: rather than enumerate dangerous constructs, each single-character atom is compiled standalone and tested against `/`. Measured results — `\D \S \W \p{P} \P{L}` and the POSIX `ascii`, `punct`, `graph`, `print` classes match `/`; `\d \s \w \p{L}`, `alpha`, `lower`, `alnum`, `digit` do not. All six known bypasses are now rejected and both real-world patterns (the EKS and Microsoft ones, each containing literal slashes) still compile and still match.

**The cost, stated plainly:** a variable-depth path can no longer be expressed. `https://example\.com/[a-z0-9/-]+` is rejected, because that class contains `/`. Per-segment patterns like `https://example\.com/realms/[a-z0-9-]+` work and are more precise, and the error message and README now say so — but an operator who genuinely needs variable depth has no way to write it.

### Not adopted, with reasons

**"An empty enumeration caches `Absent` from no evidence."** Accepted and fixed — `noAccess == len(installs)` was satisfied by `0 == 0`, so a manager disagreement that returned no installations would have cached allow-all. The positive-knowledge branch now requires `len(installs) > 0`.

**"Treat all-no-access as unknown and fail closed; do not put `Absent` in the stale cache."** The fail-open-on-`Absent` posture is deliberate, and the second half of this recommendation would cause outages for orgs that have *not* adopted the feature: no file ⇒ `Absent` cached ⇒ GitHub rate-limits ⇒ no stale entry eligible to serve ⇒ every exchange in that org denied. That is the overwhelmingly common case, so `Absent` must remain valid last-known-good. The narrow risk that remains is real and worth stating: an org that adds its *first* allowlist while GitHub is degraded can serve the previous `Absent` for up to an hour before enforcing.

One correction to my own earlier reasoning here, since the review was right to probe it: `expirable.LRU.Get` never extends a TTL, but `Add` resets it, and `cacheOrgIssuerEntry` writes `Absent` to *both* caches. So the one-hour last-known-good clock runs from the last **successful read** — including a successful `Absent` — not from first observation. Serving a stale entry does not extend it; deriving a fresh verdict does.

**"`GetAll` is a cached routing inventory, not authoritative — don't use it as security evidence."** Correct, and `Manager.GetAll`'s doc comment says so explicitly: the positive cache has no TTL, so an uninstalled App keeps appearing and a newly installed one stays hidden until a negative entry expires. After an uninstall/reinstall the stale installation can 422 while the one that could read the real policy is omitted, feeding the fail-open path. Making enumeration authoritative means invalidating on installation webhooks — a real change I did not want to smuggle into this PR. The operator-level ceiling in #1396 is the right backstop, and is one reason to want it.

**Partly addressed since.** The confirmation step above closes the half of this that was reachable: the negative cache can no longer hide a just-installed App from the allow-all decision, because that decision is now re-checked without it. The half that remains is the positive cache having no TTL, so an uninstalled App can still be enumerated and 422 — which counts as a failure and therefore fails closed, not open. Invalidating on installation webhooks is still the real fix and still out of scope here.

**"Replicas can diverge; the caches are process-local globals."** True. Two replicas can briefly enforce different modes. This matches the existing `trustPolicies`/`staleTrustPolicies` design rather than introducing a new pattern, so fixing it is a broader change than this feature.

**"No singleflight — the exchange path can stampede GitHub on cache expiry."** True, and the sharpest of the operational objections. Every primary-cache miss enumerates installations and then mints, reads, and revokes per installation, with no coordination between concurrent exchanges. `fetchTrustPolicyRaw` has the same shape, but this adds a *second* GitHub dependency to the same path and consumes the same quota. A per-org singleflight would be a contained follow-up; I did not add a new dependency at this stage.

## Follow-ups found while working, not fixed here

Each is pre-existing and out of scope; happy to file separately:

1. **`TrustPolicy.Compile` anchors patterns as `"^" + p + "$"`** (`trust_policy.go:72,86,98,108`). `|` binds looser than the anchors, so a pattern containing a top-level alternation parses as `(^A)|(B$)` and is effectively unanchored — a bypass using a pattern that reads as safe. This PR's own matcher uses `"^(?:" + p + ")$"` for exactly this reason, so the two now differ.
2. **`validatePolicies` compares `repo == ".github"` exactly** for org trust policies. An org whose `.github` repo GitHub reports as `.Github` gets its org policies parsed as repo-level `TrustPolicy` — the same class of bogus failure this PR fixes for the allowlist. The new allowlist arm folds case; the pre-existing arm was left alone.
3. **`TrustPolicy` validates the raw `token.Issuer`**, so a schemeless `iss: accounts.google.com` is rejected even though `auth.NormalizeIssuer` and go-oidc deliberately accommodate it.
4. **The stale-entry path can re-cache `Absent` without confirming it, extending the documented one-hour window by up to one primary TTL.** Serving a stale entry re-inserts it into the five-minute cache, so an `Absent` that entered both caches at `t=0` can still be served at roughly `t=64:59` — the stale entry expires at `t=60:00` but the reseeded primary outlives it. The one-hour fail-open is deliberate and documented; the extra five minutes is an artifact of reseeding. Capping the reseed at the stale entry's original expiry, or serving stale without reseeding, would fix it. **Relevant to the confirmation fix above:** an org that was previously confirmed-`Absent` and then installs an App to enable enforcement can sit on served-stale allow-all through a `GetAllFresh` outage, because the stale path runs before any confirmation. I left it alone because it is pre-existing and changing it alters behavior this PR has not been reviewed for, but it is the natural next step if the confirmation is considered worth having.
5. **`manager`'s cache pair has a lost-update race** (`ghinstall.go`). Now handled in **#1543**, which narrows the window and documents the remaining gap rather than claiming to close it. Noted here because it was found while working on this PR.

## Verification

`golangci-lint run ./...` reports zero issues, `go test ./...` passes across all packages, and `go test -race` is clean on `pkg/octosts`, `pkg/ghinstall`, and `pkg/webhook`. Every function added by the confirmation fix is at 100% statement coverage: `exhaustiveNoAccess`, `record`, `installsNotIn`, `scanInstalls`, `confirmNoAccess`, `orgIssuerLookup`. (`GetAllFresh`, `enumerate`, and `cacheKey` are also at 100%, and are covered by #1543.)

The security-critical predicates were mutation-tested rather than assumed covered — for each, the term was removed and the suite confirmed to fail: the confirmation call itself, the `installsNotIn` argument order, passing the same tally to the second pass rather than a fresh one, the "at least one answered" guard, the "GetAll reported itself complete" guard, and the negative-entry removal in `setInstalled`. One term, the redundant `!anyFailed` conjunct, is deliberately *not* detectable by mutation: it is a semantic no-op today, retained as defense-in-depth against a future edit that stops counting an unanswered installation, and its comment says exactly that.

An adversarial review pass over the change found no defect introduced by it, having attacked the two-pass tally arithmetic, installation-ID dedup and reinstall IDs, empty/subset/superset enumerations, partial and duplicate-App enumerations, 403/404/422 mint and contents classification, and the webhook path matching. Items 4 and 5 above are what it surfaced instead, both pre-existing.

One note on reading the history: `installsNotIn` is added in one commit, deleted in the next, and re-added later. That is not a mistake — it was originally introduced a commit before its caller, which fails `golangci-lint`'s `unused` check, so it was moved to land with the code that calls it. Happy to squash the review fix-ups if you would rather review a linear five-commit series.

### DIFF
--- a/pkg/octosts/org_trusted_issuers.go
+++ b/pkg/octosts/org_trusted_issuers.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"regexp/syntax"
 	"slices"
 	"strings"
 
@@ -103,6 +104,9 @@ func (c *OrgTrustedIssuers) Compile() (*IssuerAllowlist, error) {
 		if err := checkNoSlashMatchingAtom(i, p); err != nil {
 			return nil, err
 		}
+		if err := checkNoSeparatorUnderQuantifier(i, p); err != nil {
+			return nil, err
+		}
 		// Compile p ALONE first: the group below hides an unbalanced ")", so
 		// "token\.example\.com)|(" wraps to "^(?:token\.example\.com)|()$", whose "()$"
 		// alternative matches everything. A legitimate top-level "a|b" still compiles.
@@ -152,10 +156,8 @@ func (a *IssuerAllowlist) Allows(issuer string) bool {
 // "/" without containing ".", and "[.-9]" holds "/" by range. A LITERAL "/" stays allowed;
 // unterminated brackets are left to regexp.Compile.
 //
-// KNOWN LIMITATION, structural rather than a missing case: being per-ATOM, a literal "/"
-// inside a QUANTIFIED group escapes it, so "https://([a-z0-9.-]+/)*trusted\.example\.com"
-// validates yet matches "https://evil.com/x/trusted.example.com". Catching it needs AST
-// analysis; TestCompileKnownLimitationQuantifiedGroupSpansSeparators pins the gap.
+// Being per-ATOM it cannot see a "/" that only spans because of surrounding
+// STRUCTURE; checkNoSeparatorUnderQuantifier covers that half.
 func checkNoSlashMatchingAtom(i int, p string) error {
 	for j := 0; j < len(p); j++ {
 		var atom string
@@ -207,6 +209,73 @@ func checkNoSlashMatchingAtom(i int, p string) error {
 		}
 	}
 	return nil
+}
+
+// checkNoSeparatorUnderQuantifier rejects a pattern where a "/"-matching element sits
+// inside a repetition, an optional, or an alternation. The per-atom scan cannot see
+// this: every atom in "([a-z0-9.-]+/)*" is innocent on its own, yet the group spans
+// separators, so "https://([a-z0-9.-]+/)*trusted\.example\.com" matches
+// "https://evil.com/x/trusted.example.com" — attacker-chosen host, trusted name demoted
+// to a path segment. "?", "{n,m}" and alternation do the same.
+//
+// A literal "/" at a FIXED position stays allowed, so real multi-segment issuers such as
+// "https://host\.example\.com/id/[A-Z0-9]+" still validate. An author who wants
+// alternatives writes separate issuer_patterns entries instead of one alternation.
+func checkNoSeparatorUnderQuantifier(i int, p string) error {
+	re, err := syntax.Parse(p, syntax.Perl)
+	if err != nil {
+		// Unparseable: regexp.Compile reports it better than we would.
+		return nil
+	}
+	if found := separatorUnderQuantifier(re, false); found != "" {
+		return fmt.Errorf(`issuer_patterns[%d]: %s can match "/" from inside a `+
+			`repetition, optional or alternation, so the pattern can span past the `+
+			`host — "https://([a-z0-9.-]+/)*trusted\.example\.com" also permits `+
+			`"https://evil.com/x/trusted.example.com". Keep "/" at a fixed position, `+
+			`e.g. https://host\.example\.com/id/[A-Z0-9]+, and write alternatives as `+
+			`separate issuer_patterns entries rather than one alternation`, i, found)
+	}
+	return nil
+}
+
+// separatorUnderQuantifier reports the first "/"-matching element reachable under a
+// variable-length or alternating node, or "" if there is none. variable becomes true
+// once inside such a node and stays true for the whole subtree.
+func separatorUnderQuantifier(re *syntax.Regexp, variable bool) string {
+	switch re.Op {
+	case syntax.OpStar, syntax.OpPlus, syntax.OpQuest, syntax.OpRepeat, syntax.OpAlternate:
+		variable = true
+	}
+	if variable {
+		if what := slashMatchingElement(re); what != "" {
+			return what
+		}
+	}
+	for _, sub := range re.Sub {
+		if found := separatorUnderQuantifier(sub, variable); found != "" {
+			return found
+		}
+	}
+	return ""
+}
+
+// slashMatchingElement names re if it can match "/" on its own, else "".
+func slashMatchingElement(re *syntax.Regexp) string {
+	switch re.Op {
+	case syntax.OpLiteral:
+		if slices.Contains(re.Rune, '/') {
+			return `a literal "/"`
+		}
+	case syntax.OpCharClass:
+		for j := 0; j+1 < len(re.Rune); j += 2 {
+			if re.Rune[j] <= '/' && '/' <= re.Rune[j+1] {
+				return "a character class containing \"/\""
+			}
+		}
+	case syntax.OpAnyChar, syntax.OpAnyCharNotNL:
+		return `"." (which matches "/")`
+	}
+	return ""
 }
 
 // endOfEscape returns the index just past the character-class escape starting at the
--- a/pkg/octosts/org_trusted_issuers.go
+++ b/pkg/octosts/org_trusted_issuers.go
@@ -100,11 +100,7 @@ func (c *OrgTrustedIssuers) Compile() (*IssuerAllowlist, error) {
 		if len(p) > maxPatternLen {
 			return nil, fmt.Errorf("issuer_patterns[%d] too long: %d chars (max %d)", i, len(p), maxPatternLen)
 		}
-		// First, so a loose-and-malformed pattern reports its looseness.
-		if err := checkNoSlashMatchingAtom(i, p); err != nil {
-			return nil, err
-		}
-		if err := checkNoSeparatorUnderQuantifier(i, p); err != nil {
+		if err := checkPatternCannotSpanHost(i, p); err != nil {
 			return nil, err
 		}
 		// Compile p ALONE first: the group below hides an unbalanced ")", so
@@ -146,224 +142,67 @@ func (a *IssuerAllowlist) Allows(issuer string) bool {
 	return false
 }
 
-// checkNoSlashMatchingAtom rejects any single-character atom that can match "/": such an
-// atom lets a pattern span past the host, so "https://.*\.example\.com" would also permit
-// "https://totally-evil.com/x.example.com". BEST-EFFORT FOOTGUN GUARD, NOT A SECURITY
-// BOUNDARY — the author is the org owner.
+// checkPatternCannotSpanHost rejects an issuer_pattern that can match "/" and so span
+// past the host: "https://.*\.example\.com" also permits "https://evil.com/x.example.com".
+// BEST-EFFORT FOOTGUN GUARD, NOT A SECURITY BOUNDARY — the author is the org owner.
 //
-// The test is EMPIRICAL — each atom compiled alone as "^atom$" and asked whether it matches
-// "/" — because banned spellings cannot be enumerated: "\S", "[^\n]" and "[[:ascii:]]" match
-// "/" without containing ".", and "[.-9]" holds "/" by range. A LITERAL "/" stays allowed;
-// unterminated brackets are left to regexp.Compile.
-//
-// Being per-ATOM it cannot see a "/" that only spans because of surrounding
-// STRUCTURE; checkNoSeparatorUnderQuantifier covers that half.
-func checkNoSlashMatchingAtom(i int, p string) error {
-	for j := 0; j < len(p); j++ {
-		var atom string
-
-		switch p[j] {
-		case '.':
-			atom = "."
-
-		case '\\':
-			end := endOfEscape(p, j)
-			if end < 0 {
-				// Skip the escaped byte so "\." is never read as a wildcard.
-				j++
-				continue
-			}
-			atom, j = p[j:end], end-1
-
-		case '[':
-			end := endOfBracketExpression(p, j)
-			if end < 0 {
-				return nil
-			}
-			atom, j = p[j:end], end-1
-
-		default:
-			continue
-		}
-
-		re, err := regexp.Compile("^" + atom + "$")
-		if err != nil {
-			// Fail CLOSED: an atom we cannot analyze is not one we know is safe, and
-			// skipping it would turn every scanner mis-parse into a silent bypass.
-			return fmt.Errorf(`issuer_patterns[%d]: cannot verify that %q cannot `+
-				`match "/" (%v), so the pattern is rejected rather than assumed `+
-				`safe — rewrite it with a plain class such as [a-z0-9-]`, i, atom, err)
-		}
-		if re.MatchString("/") {
-			return fmt.Errorf(`issuer_patterns[%d]: %q can match "/", so the pattern `+
-				`can span past the host into an attacker-controlled domain, as `+
-				`"https://.*\.example\.com" also permits `+
-				`"https://totally-evil.com/x.example.com" — escape literal dots as `+
-				`"\." and bound each varying part, e.g. `+
-				`https://[a-z0-9-]+\.example\.com. Write a path one segment at a `+
-				`time with literal separators, e.g. `+
-				`https://example\.com/realms/[a-z0-9-]+, because a class that `+
-				`itself contains "/" is rejected for the same reason. `+
-				`This is a best-effort check, `+
-				`not a guarantee: write patterns as narrowly as you can`, i, atom)
-		}
-	}
-	return nil
-}
-
-// checkNoSeparatorUnderQuantifier rejects a pattern where a "/"-matching element sits
-// inside a repetition, an optional, or an alternation. The per-atom scan cannot see
-// this: every atom in "([a-z0-9.-]+/)*" is innocent on its own, yet the group spans
-// separators, so "https://([a-z0-9.-]+/)*trusted\.example\.com" matches
-// "https://evil.com/x/trusted.example.com" — attacker-chosen host, trusted name demoted
-// to a path segment. "?", "{n,m}" and alternation do the same.
-//
-// A literal "/" at a FIXED position stays allowed, so real multi-segment issuers such as
-// "https://host\.example\.com/id/[A-Z0-9]+" still validate. An author who wants
-// alternatives writes separate issuer_patterns entries instead of one alternation.
-func checkNoSeparatorUnderQuantifier(i int, p string) error {
+// It inspects the regexp/syntax AST — the same parse regexp.Compile performs — so it
+// cannot disagree with the matcher about tokenization. A hand-written byte scanner did
+// this before and disagreed six times, each a bypass.
+func checkPatternCannotSpanHost(i int, p string) error {
 	re, err := syntax.Parse(p, syntax.Perl)
 	if err != nil {
-		// Unparseable: regexp.Compile reports it better than we would.
+		// Unparseable: the compile below reports it better than we would.
 		return nil
 	}
-	if found := separatorUnderQuantifier(re, false); found != "" {
-		return fmt.Errorf(`issuer_patterns[%d]: %s can match "/" from inside a `+
-			`repetition, optional or alternation, so the pattern can span past the `+
-			`host — "https://([a-z0-9.-]+/)*trusted\.example\.com" also permits `+
-			`"https://evil.com/x/trusted.example.com". Keep "/" at a fixed position, `+
-			`e.g. https://host\.example\.com/id/[A-Z0-9]+, and write alternatives as `+
-			`separate issuer_patterns entries rather than one alternation`, i, found)
+	if what := spanningElement(re, false); what != "" {
+		return fmt.Errorf(`issuer_patterns[%d]: %s, so the pattern can span past the `+
+			`host into an attacker-controlled domain, as "https://.*\.example\.com" also `+
+			`permits "https://totally-evil.com/x.example.com" — escape literal dots as `+
+			`"\." and bound each varying part, e.g. https://[a-z0-9-]+\.example\.com. Keep `+
+			`"/" at a fixed position, writing a path one segment at a time, and give `+
+			`alternatives their own issuer_patterns entries. This is a best-effort check, `+
+			`not a guarantee: write patterns as narrowly as you can`, i, what)
 	}
 	return nil
 }
 
-// separatorUnderQuantifier reports the first "/"-matching element reachable under a
-// variable-length or alternating node, or "" if there is none. variable becomes true
-// once inside such a node and stays true for the whole subtree.
-func separatorUnderQuantifier(re *syntax.Regexp, variable bool) string {
+// spanningElement describes the first element that lets the pattern match "/", or "".
+//
+// A class or "." matching "/" is rejected ANYWHERE: parsing desugars "\S", "[^\n]",
+// "[[:ascii:]]", "\p{P}" and "[.-9]" into rune ranges, so one range test covers every
+// spelling. A literal "/" is rejected only under a repetition, optional or alternation
+// — at a fixed position it is how real multi-segment issuers are written, e.g.
+// https://host\.example\.com/id/[A-Z0-9]+. variable carries that down the subtree.
+func spanningElement(re *syntax.Regexp, variable bool) string {
 	switch re.Op {
 	case syntax.OpStar, syntax.OpPlus, syntax.OpQuest, syntax.OpRepeat, syntax.OpAlternate:
 		variable = true
 	}
-	if variable {
-		if what := slashMatchingElement(re); what != "" {
-			return what
-		}
-	}
-	for _, sub := range re.Sub {
-		if found := separatorUnderQuantifier(sub, variable); found != "" {
-			return found
-		}
-	}
-	return ""
-}
 
-// slashMatchingElement names re if it can match "/" on its own, else "".
-func slashMatchingElement(re *syntax.Regexp) string {
 	switch re.Op {
-	case syntax.OpLiteral:
-		if slices.Contains(re.Rune, '/') {
-			return `a literal "/"`
-		}
+	case syntax.OpAnyChar, syntax.OpAnyCharNotNL:
+		return `"." can match "/"`
+
 	case syntax.OpCharClass:
 		for j := 0; j+1 < len(re.Rune); j += 2 {
 			if re.Rune[j] <= '/' && '/' <= re.Rune[j+1] {
-				return "a character class containing \"/\""
+				return `a character class can match "/"`
 			}
 		}
-	case syntax.OpAnyChar, syntax.OpAnyCharNotNL:
-		return `"." (which matches "/")`
+
+	case syntax.OpLiteral:
+		if variable && slices.Contains(re.Rune, '/') {
+			return `a literal "/" sits inside a repetition, optional or alternation`
+		}
+	}
+
+	for _, sub := range re.Sub {
+		if what := spanningElement(sub, variable); what != "" {
+			return what
+		}
 	}
 	return ""
-}
-
-// endOfEscape returns the index just past the character-class escape starting at the
-// backslash at p[j], or -1 when there is no atom to test (an escaped literal or
-// trailing backslash) and the caller should skip two bytes.
-func endOfEscape(p string, j int) int {
-	if j+1 >= len(p) {
-		return -1
-	}
-	switch p[j+1] {
-	case 'd', 'D', 's', 'S', 'w', 'W':
-		return j + 2
-
-	// Numeric escapes SPELL characters rather than naming classes ("\x2F", "\x{2F}" and
-	// "\057" are all "/"); as escaped literals they would be skipped and never tested.
-	case 'x':
-		if j+2 >= len(p) {
-			return -1
-		}
-		if p[j+2] == '{' {
-			if k := strings.IndexByte(p[j+2:], '}'); k >= 0 {
-				return j + 2 + k + 1
-			}
-			return -1
-		}
-		if j+3 < len(p) && isHexDigit(p[j+2]) && isHexDigit(p[j+3]) {
-			return j + 4
-		}
-		return -1
-	case '0', '1', '2', '3', '4', '5', '6', '7':
-		end := j + 2
-		for end < len(p) && end < j+4 && p[end] >= '0' && p[end] <= '7' {
-			end++
-		}
-		return end
-
-	case 'p', 'P':
-		if j+2 >= len(p) {
-			return -1
-		}
-		if p[j+2] != '{' {
-			return j + 3
-		}
-		if k := strings.IndexByte(p[j+2:], '}'); k >= 0 {
-			return j + 2 + k + 1
-		}
-		return -1
-	default:
-		return -1
-	}
-}
-
-func isHexDigit(b byte) bool {
-	return b >= '0' && b <= '9' || b >= 'a' && b <= 'f' || b >= 'A' && b <= 'F'
-}
-
-// endOfBracketExpression returns the index just past the bracket expression opening at
-// p[j], or -1 if unterminated. More than a search for "]": a backslash escapes the next
-// byte ("[a-z\]]"), a POSIX class carries its own "]" ("[[:alpha:]]"), and a leading "]"
-// is a literal member ("[^]]").
-func endOfBracketExpression(p string, j int) int {
-	k := j + 1
-	if k < len(p) && p[k] == '^' {
-		k++
-	}
-	// RE2 reads a first-position "]" as a literal; stopping there would extract the
-	// uncompilable "[^]" and leave the real class untested.
-	if k < len(p) && p[k] == ']' {
-		k++
-	}
-	for ; k < len(p); k++ {
-		switch p[k] {
-		case '\\':
-			k++
-		case '[':
-			// "[:name:]" is a POSIX class only when a ":]" actually follows; otherwise
-			// RE2 reads this "[" as an ordinary member and the real "]" is still ahead,
-			// so keep scanning: bailing out let "[[:/-~]", holding "/", pass untested.
-			if strings.HasPrefix(p[k:], "[:") {
-				if end := strings.Index(p[k:], ":]"); end >= 0 {
-					k += end + 1
-				}
-			}
-		case ']':
-			return k + 1
-		}
-	}
-	return -1
 }
 
 // checkLowercaseSchemeAndHost rejects uppercase in an issuer's scheme or host: matching
--- a/pkg/octosts/org_trusted_issuers.go
+++ b/pkg/octosts/org_trusted_issuers.go
@@ -1,0 +1,315 @@
+// Copyright 2026 Chainguard, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package octosts
+
+import (
+	"errors"
+	"fmt"
+	"regexp"
+	"slices"
+	"strings"
+
+	"github.com/octo-sts/app/pkg/oidcvalidate"
+	"sigs.k8s.io/yaml"
+)
+
+// Mode selects how an org's trusted-issuer allowlist is applied.
+type Mode string
+
+const (
+	// ModeEnforce rejects tokens whose issuer is not permitted. Default.
+	ModeEnforce Mode = "enforce"
+	// ModeAudit permits them and records what would have been rejected.
+	ModeAudit Mode = "audit"
+)
+
+// Size caps bound memory and compile time; RE2 cannot backtrack catastrophically.
+const (
+	maxIssuers        = 64
+	maxIssuerPatterns = 16
+	maxPatternLen     = 256
+)
+
+// OrgTrustedIssuers is an org-wide allowlist of OIDC issuers read from
+// OrgTrustedIssuersPath. When present, only issuers it permits may federate.
+type OrgTrustedIssuers struct {
+	Mode Mode `json:"mode,omitempty" jsonschema:"enum=enforce,enum=audit"`
+
+	// Issuers is matched byte-exactly.
+	Issuers []string `json:"issuers,omitempty"`
+
+	// IssuerPatterns are anchored regexps matching the entire issuer URL.
+	IssuerPatterns []string `json:"issuer_patterns,omitempty"`
+}
+
+// ParseOrgTrustedIssuers unmarshals and compiles raw file contents. Both the
+// exchange path and the webhook check call this, so they cannot disagree.
+func ParseOrgTrustedIssuers(raw []byte) (*IssuerAllowlist, error) {
+	var cfg OrgTrustedIssuers
+	if err := yaml.UnmarshalStrict(raw, &cfg); err != nil {
+		return nil, fmt.Errorf("parsing organization trusted issuers: %w", err)
+	}
+	return cfg.Compile()
+}
+
+// IssuerAllowlist is a compiled OrgTrustedIssuers, immutable and concurrent-safe.
+type IssuerAllowlist struct {
+	mode     Mode
+	issuers  []string
+	patterns []*regexp.Regexp
+}
+
+// Compile validates the configuration and prepares it for matching. It returns a
+// new value, unlike TrustPolicy.Compile, so a cached allowlist cannot be
+// invalidated by a concurrent recompile.
+func (c *OrgTrustedIssuers) Compile() (*IssuerAllowlist, error) {
+	mode := c.Mode
+	switch mode {
+	case "":
+		mode = ModeEnforce
+	case ModeEnforce, ModeAudit:
+	default:
+		return nil, fmt.Errorf("unknown mode %q (want %q or %q)", c.Mode, ModeEnforce, ModeAudit)
+	}
+
+	// An empty file would deny all federation in the org: a silent outage.
+	if len(c.Issuers) == 0 && len(c.IssuerPatterns) == 0 {
+		return nil, errors.New("must list at least one entry in issuers or issuer_patterns")
+	}
+
+	if len(c.Issuers) > maxIssuers {
+		return nil, fmt.Errorf("too many issuers: %d (max %d)", len(c.Issuers), maxIssuers)
+	}
+	if len(c.IssuerPatterns) > maxIssuerPatterns {
+		return nil, fmt.Errorf("too many issuer_patterns: %d (max %d)", len(c.IssuerPatterns), maxIssuerPatterns)
+	}
+
+	for _, iss := range c.Issuers {
+		if !oidcvalidate.IsValidIssuer(iss) {
+			return nil, fmt.Errorf("invalid issuer %q", iss)
+		}
+		if err := checkLowercaseSchemeAndHost(iss); err != nil {
+			return nil, err
+		}
+	}
+
+	patterns := make([]*regexp.Regexp, 0, len(c.IssuerPatterns))
+	for i, p := range c.IssuerPatterns {
+		if len(p) > maxPatternLen {
+			return nil, fmt.Errorf("issuer_patterns[%d] too long: %d chars (max %d)", i, len(p), maxPatternLen)
+		}
+		// Before compiling: a loose-and-malformed pattern should report looseness.
+		if err := checkNoSlashMatchingAtom(i, p); err != nil {
+			return nil, err
+		}
+		// Compile p ALONE first: the group below hides an unbalanced ")", so
+		// "token\.example\.com)|(" wraps to "^(?:token\.example\.com)|()$", whose
+		// "()$" alternative matches everything — allow-all, reported Valid by the
+		// webhook. A legitimate top-level "a|b" still compiles alone.
+		if _, err := regexp.Compile(p); err != nil {
+			return nil, fmt.Errorf("invalid issuer_pattern %q: %w", p, err)
+		}
+		// Non-capturing group required: "|" binds looser than the anchors, so
+		// "^"+p+"$" parses as "(^A)|(B$)", leaving an alternation unanchored.
+		re, err := regexp.Compile("^(?:" + p + ")$")
+		if err != nil {
+			return nil, fmt.Errorf("invalid issuer_pattern %q: %w", p, err)
+		}
+		patterns = append(patterns, re)
+	}
+
+	return &IssuerAllowlist{
+		mode:     mode,
+		issuers:  slices.Clone(c.Issuers),
+		patterns: patterns,
+	}, nil
+}
+
+func (a *IssuerAllowlist) Mode() Mode { return a.mode }
+
+// Allows reports whether issuer is permitted. The two lists are a union,
+// deliberately unlike TrustPolicy where they are mutually exclusive. Patterns
+// honour inline flags such as (?i), so matching is only as strict as the pattern.
+func (a *IssuerAllowlist) Allows(issuer string) bool {
+	if slices.Contains(a.issuers, issuer) {
+		return true
+	}
+	for _, re := range a.patterns {
+		if re.MatchString(issuer) {
+			return true
+		}
+	}
+	return false
+}
+
+// checkNoSlashMatchingAtom rejects an issuer_pattern containing any
+// single-character atom that can match "/". BEST-EFFORT FOOTGUN GUARD, NOT A
+// SECURITY BOUNDARY — the author is the org owner. The accident it catches: an
+// atom matching "/" lets a pattern span past the host, so
+// "https://.*\.example\.com" also permits "https://totally-evil.com/x.example.com".
+//
+// The test is EMPIRICAL — each atom compiled alone as "^atom$", asked whether it
+// matches "/" — because banned spellings cannot be enumerated: "\S", "[^\n]" and
+// "[[:ascii:]]" match "/" without containing ".", and "[.-9]" holds "/" by range.
+// A LITERAL "/" stays allowed. Unterminated brackets are left to regexp.Compile.
+func checkNoSlashMatchingAtom(i int, p string) error {
+	for j := 0; j < len(p); j++ {
+		var atom string
+
+		switch p[j] {
+		case '.':
+			atom = "."
+
+		case '\\':
+			end := endOfEscape(p, j)
+			if end < 0 {
+				// Skip the escaped byte so "\." is never read as a wildcard.
+				j++
+				continue
+			}
+			atom, j = p[j:end], end-1
+
+		case '[':
+			end := endOfBracketExpression(p, j)
+			if end < 0 {
+				return nil
+			}
+			atom, j = p[j:end], end-1
+
+		default:
+			continue
+		}
+
+		re, err := regexp.Compile("^" + atom + "$")
+		if err != nil {
+			// Fail CLOSED: an atom we cannot analyze is not an atom we know is safe.
+			// Skipping it would turn every scanner mis-parse into a silent bypass.
+			return fmt.Errorf(`issuer_patterns[%d]: cannot verify that %q cannot `+
+				`match "/" (%v), so the pattern is rejected rather than assumed `+
+				`safe — rewrite it with a plain class such as [a-z0-9-]`, i, atom, err)
+		}
+		if re.MatchString("/") {
+			return fmt.Errorf(`issuer_patterns[%d]: %q can match "/", so the pattern `+
+				`can span past the host into an attacker-controlled domain, as `+
+				`"https://.*\.example\.com" also permits `+
+				`"https://totally-evil.com/x.example.com" — escape literal dots as `+
+				`"\." and bound each varying part, e.g. `+
+				`https://[a-z0-9-]+\.example\.com. Write a path one segment at a `+
+				`time with literal separators, e.g. `+
+				`https://example\.com/realms/[a-z0-9-]+, because a class that `+
+				`itself contains "/" is rejected for the same reason. `+
+				`This is a best-effort check, `+
+				`not a guarantee: write patterns as narrowly as you can`, i, atom)
+		}
+	}
+	return nil
+}
+
+// endOfEscape returns the index just past the character-class escape starting at
+// the backslash at p[j], or -1 when there is no atom to test (an escaped literal
+// or trailing backslash), meaning the caller should skip two bytes.
+func endOfEscape(p string, j int) int {
+	if j+1 >= len(p) {
+		return -1
+	}
+	switch p[j+1] {
+	case 'd', 'D', 's', 'S', 'w', 'W':
+		return j + 2
+
+	// Numeric escapes SPELL characters rather than naming classes: "\x2F",
+	// "\x{2F}" and "\057" are all "/". Treated as escaped literals they would be
+	// skipped and the slash they encode would never be tested.
+	case 'x':
+		if j+2 >= len(p) {
+			return -1
+		}
+		if p[j+2] == '{' {
+			if k := strings.IndexByte(p[j+2:], '}'); k >= 0 {
+				return j + 2 + k + 1
+			}
+			return -1
+		}
+		if j+3 < len(p) && isHexDigit(p[j+2]) && isHexDigit(p[j+3]) {
+			return j + 4
+		}
+		return -1
+	case '0', '1', '2', '3', '4', '5', '6', '7':
+		end := j + 2
+		for end < len(p) && end < j+4 && p[end] >= '0' && p[end] <= '7' {
+			end++
+		}
+		return end
+
+	case 'p', 'P':
+		if j+2 >= len(p) {
+			return -1
+		}
+		if p[j+2] != '{' {
+			return j + 3
+		}
+		if k := strings.IndexByte(p[j+2:], '}'); k >= 0 {
+			return j + 2 + k + 1
+		}
+		return -1
+	default:
+		return -1
+	}
+}
+
+func isHexDigit(b byte) bool {
+	return b >= '0' && b <= '9' || b >= 'a' && b <= 'f' || b >= 'A' && b <= 'F'
+}
+
+// endOfBracketExpression returns the index just past the bracket expression
+// opening at p[j], or -1 if unterminated. More than a search for "]": a
+// backslash escapes the next byte ("[a-z\]]"), a POSIX class carries its own "]"
+// ("[[:alpha:]]"), and a leading "]" is a literal member ("[^]]").
+func endOfBracketExpression(p string, j int) int {
+	k := j + 1
+	if k < len(p) && p[k] == '^' {
+		k++
+	}
+	// RE2 reads a first-position "]" as a literal. Stopping on it would extract
+	// the uncompilable "[^]" and leave the real class untested.
+	if k < len(p) && p[k] == ']' {
+		k++
+	}
+	for ; k < len(p); k++ {
+		switch p[k] {
+		case '\\':
+			k++
+		case '[':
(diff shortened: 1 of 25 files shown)


jmeridth:
:robot: AI-assisted review.

Thank you for this, @salekseev, and apologies for the long delay in getting to it. There have been substantial changes to the codebase since you opened this, so it no longer merges cleanly. Could you rebase onto `main` and resolve the conflicts? Once it's updated we'll review promptly. Thanks again for contributing.

jmeridth:
@salekseev thanks. Taking a look. 4K lines added is a bit brutal but 60% of it is tests. We do need to drop the `.vscode/settings.json` though. I am going to have to take you up on your offer to split this. It is a great PR but splitting it up will be necessary IMO. You good with that?

salekseev:
Yes, absolutely — splitting is the right call and I should have led with it. Here's the plan; tell me if you'd rather slice it differently.

**Four independent PRs off `main`, each mergeable on its own, then #901 becomes just the feature:**

1. **Rate-limit classification fix** — the pre-existing bug. `fetchTrustPolicyRaw` matched `*github.ErrorResponse`, but a real primary rate limit yields `*github.RateLimitError`, which doesn't unwrap to it, so stale-cache fallback and cross-app retry were both dead. Adds the `X-RateLimit-Remaining` header to the fake first (breaking four tests), then fixes the classifier. ~3 files, no feature dependency.

2. **`newGitHubClient` consolidation** — `pkg/octosts` had two copies of the `opts`/`WithEnterpriseURLs`/`NewClient` block and `pkg/ghinstall`/`pkg/webhook` have their own. Collapses the `pkg/octosts` copies into one helper, with a test pinning the base URL for github.com and GHES, and a test enforcing it's the package's only constructor. Pure refactor of existing duplication; the GHES bug I mentioned was caused by a fourth copy drifting.

3. **Webhook path-predicate consolidation + the unfiltered directory scan** — five inlined `filepath.Match` calls become one `isValidatedPath`, and `handleCheckSuite`'s new-branch/initial-commit branch gets filtered like every other path. **This one contains the behavior change:** a non-policy file in `.github/chainguard` (a `README.md`, say) was previously fetched and parsed as a `TrustPolicy` on that path, and now isn't.

4. **`ghinstall.Manager.GetAll`** — exhaustive installation enumeration, with the contract that a non-nil error means the result is *not* exhaustive even when non-empty. `Get` is a routing primitive (`pickByQuota` is argmax with no rotation), so it can't be used to enumerate. No consumer until 5, so happy to fold this into 5 instead if you'd rather not merge an unused method.

5. **#901** — the feature itself, rebased to drop the above as they merge. Should end up well under half its current size.

Heads-up that 2 and 3 won't be clean cherry-picks: my commits for them mix in feature-dependent bits (the exported path constant, a `Revoke` call in a file that doesn't exist yet), so I'll hand-craft the standalone versions rather than pretend they split cleanly. Each will get its own description.

---

**On `.vscode/settings.json` — happy to drop it, but I want to check what you actually want, because the file predates this PR and the README points at it.**

The file exists on `main` today with:

```json
"./pkg/octosts/octosts.TrustPolicy.json": "/.github/chainguard/*"
```

That glob matches *everything* in that directory. So once `trusted-token-issuers.yaml` exists, an editor with [vscode-yaml](https://github.com/redhat-developer/vscode-yaml) — which README:76 recommends — will validate the allowlist against the **TrustPolicy** schema and show errors on a perfectly valid file. My change was two lines: narrow that glob to `*.sts.yaml`, and map the new schema.

So which would you prefer?

- **Keep only the glob narrowing** (one line, `*` → `*.sts.yaml`), drop the new mapping. Fixes the mis-validation without adding anything.
- **Drop the file change entirely** and document the mapping in the README so users add it to their own settings. Accepts wrong-schema errors for anyone using the committed config.
- **Remove `.vscode/settings.json` from the repo** and move all of it to README guidance, if the objection is to committing editor config at all. Happy to do that as its own PR.
- **Keep both lines** as-is.

Related: is the `#### Autocomplete` section in the README the way you want new schemas surfaced? It links schema files directly, so a new schema wants an entry there regardless of what happens to the vscode config — I've added one, and I can drop or reshape it to match whatever you prefer.

salekseev:
Split done — four PRs, each green standalone against `main`, each with an issue describing the problem it solves:

| PR | Issue | What |
|---|---|---|
| #1524 | #1526 | Rate-limit misclassification. Pre-existing bug: a real primary rate limit yields `*github.RateLimitError`, which doesn't unwrap to `*github.ErrorResponse`, so the stale-cache fallback and cross-app retry were both dead in production. |
| #1529 | #1528 | One GitHub client constructor in `pkg/octosts`. Pre-existing duplication; a fourth copy that drifted is what lost the GHES base URL. |
| #1531 | #1530 | Webhook path predicate + the unfiltered directory scan. **Contains a behaviour change** — non-policy files in `.github/chainguard` stop being validated on the initial-commit path. |
| #1525 | #1527 | `Manager.GetAll`. `Get` is a routing primitive and can't enumerate. Has no in-tree caller until #901 — happy to fold it back if you'd rather not merge an unused method. |

**Suggested order: #1524, #1529, #1531, #1525, then #901.** The first three are pre-existing bugs and refactors that stand on their own merits; #1525 is the only one that exists for #901's benefit.

One thing I couldn't do: these aren't stacked. A PR's base must be a branch in this repo and I can't push branches here, so each targets `main` and #901 still *contains* their commits until they merge — its diff won't shrink until then. If you'd prefer to create base branches upstream so we can stack properly, I'm happy to redo it that way.

#901 itself is rewritten to describe only the feature, with the split table, the failure-mode table, and the reasoning behind the fail-open decisions. Two adversarial review passes are summarised there too, including the one that caught my first `issuer_patterns` guard being incomplete.

Still waiting on your call re: `.vscode/settings.json` — the question is in #901's description under "Open question". Short version: `main`'s existing glob is `/.github/chainguard/*`, which matches everything, so once the allowlist file exists an editor validates it against the **TrustPolicy** schema and reports errors on a valid file. Four options listed there; happy with any of them, including removing the file from the repo entirely.

jmeridth:
:robot: All four split-out prerequisites are now merged: #1524 (rate-limit classification), #1525 (`Manager.GetAll`), #1529 (one client constructor), and #1531 (validated-path filter). Thanks for splitting it up so cleanly -- it made each piece easy to reason about, and they're all solid.

Could you rebase this onto current `main`? It should shrink to just the org-trusted-issuers feature now that the enumeration primitive, the GHES-safe client helper, the path filter, and the rate-limit fix all live in `main`. Once it's rebased we'll do the full review.

One thing to carry into that review (from the #1525 discussion): `GetAll` is exhaustive only over *observed* state -- it can return `(empty, nil)` for up to the 5-minute negative-cache TTL after a fresh install. Worth making sure the trusted-issuer lookup treats "nil error + empty" as "none observed" (not "none exist") and fails per-installation read errors closed, so the enumeration can't read as "no restriction" during that window.

jmeridth:
:robot: (high) AI-assisted review -- 4 models (Opus 4.8, Sonnet 5, Haiku 4.5, Fable 5); both findings independently reproduced. (Anchored inline on the two relevant lines.)

First -- this is a big step up from the original. Verified sound: fail-closed on config/parse/enumeration errors, `mode` defaults to `enforce`, empty-file errors, exact-match is byte-exact against the *verified* issuer, the `total>0` empty-enumeration guard, immutable `IssuerAllowlist` (`-race` clean), the exchange checking after signature verification, and the webhook sharing `ParseOrgTrustedIssuers` (composes cleanly with #1531). Two things block:

**HIGH -- org-issuer enforcement can silently disable in a multi-app deployment (`GetAll` truncation).**
`orgIssuerLookup` concludes `Absent` (allow-all) when `enumErr == nil && exhaustiveNoAccess()`. But `Manager.GetAll` is exhaustive only over *observed* state: a negative-cached installation returns `manager.GetAll -> (nil, nil)`, so `roundRobin.GetAll` drops it **with no error** -- `enumErr == nil` does not mean the enumeration was complete.
Scenario (multi-app, incl. the public deployment): App A is installed but its selected-repos exclude `.github` (blind); App B -- which holds `.github` and the real `trusted-token-issuers.yaml` -- gets negative-cached for the owner via routine quota polling (`pickByQuota` calls `Get` on every manager each request), then is installed. For the negative-cache window (and the resulting `Absent` is cached ~5 min in `orgIssuers` + ~1 h in `staleOrgIssuers`), `GetAll` returns only App A -> `total=1, noAccess=1, anyFailed=false` -> `exhaustiveNoAccess()==true` -> `Absent` -> enforcement off org-wide. A repo-trust-policy writer (the adversary this feature targets) can, in that window, point a policy at an attacker-controlled issuer and exchange tokens.
The `s.rrm.Get`-before-`checkOrgTrustedIssuers` ordering only protects the installation `Get` resolves to (App A, which *is* installed) -- it can't protect the managers `GetAll` silently skips. Reproduced with real `ghinstall.NewRoundRobin` + two fake App servers (the `enumMgr` test double bypasses this caching layer, which is why the suite stays green).
Fix direction: the org-issuer enumeration can't treat `enumErr == nil` as "complete" given `GetAll`'s observed-only contract. Either bypass the negative cache for this security enumeration (force a fresh `ListInstallations` per app), or treat "app absent from the `GetAll` result" as *unknown* (fail-closed / stale) rather than *not-installed* when deriving `exhaustiveNoAccess`. A regression test built on real `ghinstall` managers (not `enumMgr`) would lock it in.

**MEDIUM -- `issuer_patterns` anchor guard bypassable with unbalanced parens.**
The `^(?:` + p + `)$` wrapper stops a bare top-level `|`, but an unbalanced `)` in `p` closes the group early. `issuer_patterns: ["token\.example\.com)|("]` compiles to `^(?:token\.example\.com)|()$`, whose `()$` alternative matches every string -> allow-all. It passes `checkNoSlashMatchingAtom` (no atom to test) and `regexp.Compile` (parens balance), and since the webhook shares the parser it's reported **Valid** -- false assurance for a guard that explicitly claims to stop "a pattern that reads as safe." `TestAnchoringRejectsAlternationBypass` only covers balanced `a|b`.
Fix: reject unbalanced parens (or a top-level `|` outside the wrapper) before compiling; a regression test with `token\.example\.com)|(` pins it.

Two LOW, documented/bounded (not blocking): stale-`Absent` served under GitHub API duress on first-enablement; a removed issuer lingers up to the ~1 h stale TTL after a config typo.

Happy to help think through the HIGH -- it's the load-bearing consequence of `GetAll`'s observed-only contract we flagged on #1525. Genuinely strong work otherwise.

pkg/octosts/org_trusted_issuers_store.go
```diff
@@ -0,0 +1,528 @@
...
+
+	var t orgIssuerTally
+	if entry, ok := s.scanInstalls(ctx, owner, installs, &t); ok {
+		return s.settleOrgIssuerEntry(ctx, owner, entry), nil
+	}
+
+	// Absent requires POSITIVE knowledge: every installation agreed it cannot see
+	// .github. A failed or partial enumeration is not evidence of absence and is
+	// never cached. See exhaustiveNoAccess.
+	if enumErr == nil && t.exhaustiveNoAccess() {
```
jmeridth, line 446: :robot: (high) This `enumErr == nil` gate treats the `GetAll` result as a complete enumeration, but `GetAll` is exhaustive only over *observed* state -- a negative-cached installation returns `manager.GetAll -> (nil, nil)` and is silently dropped from `roundRobin.GetAll` with **no error**. So in a multi-app deployment, a `.github`-blind App A plus a negative-cache-invisible App B (which holds the allowlist) makes `exhaustiveNoAccess()` true off App A alone -> `Absent` -> allow-all, disabling org-wide enforcement for the negative-cache/stale window. Reproduced with real `ghinstall.NewRoundRobin`. Fix: don't treat `enumErr == nil` as complete here (bypass the negative cache for this enumeration, or treat an app missing from the result as unknown/fail-closed). Full write-up + repro in the top-level review.

pkg/octosts/org_trusted_issuers.go
```diff
@@ -0,0 +1,335 @@
...
+			return nil, fmt.Errorf("issuer_patterns[%d] too long: %d chars (max %d)", i, len(p), maxPatternLen)
+		}
+		// Before compiling: a loose-and-malformed pattern should report looseness.
+		if err := checkNoSlashMatchingAtom(i, p); err != nil {
+			return nil, err
+		}
+		// Non-capturing group required: "|" binds looser than the anchors, so
+		// "^"+p+"$" would parse as "(^A)|(B$)" and leave an alternation pattern
+		// effectively unanchored — a bypass using a pattern that reads as safe.
+		re, err := regexp.Compile("^(?:" + p + ")$")
```
jmeridth, line 114: :robot: (medium) An unbalanced `)` in `p` closes this `(?:` group early, re-opening the exact alternation bypass the comment above says it prevents. `p = token\.example\.com)|(` compiles to `^(?:token\.example\.com)|()$`, whose `()$` alternative matches every string -> allow-all. It passes `checkNoSlashMatchingAtom` (no atom to test) and `regexp.Compile` (parens balance), and the webhook reports it **Valid**. Reject unbalanced parens / a top-level `|` before compiling (a standalone `regexp.Compile(p)` first, or a paren-balance check); regression-test with `token\.example\.com)|(`. Full detail in the top-level review.

salekseev: Confirmed and fixed in `1f02e9a`. `Allows()` returned true for every issuer I tried, including the empty string — a total bypass, and the webhook did report the file Valid.

Fixed by compiling `p` on its own before wrapping it, which is a syntax error for an unbalanced `)` while still accepting a legitimate top-level `a|b` that the group anchors correctly. I preferred that to a paren-balance check because it delegates to the parser rather than reimplementing part of it.

Regression test is `TestCompileRejectsPatternsThatEscapeTheAnchoringGroup`, with your `token\.example\.com)|(` as a row. It asserts the *wrapped* form still compiles first, so the row fails loudly rather than silently passing if Go ever starts rejecting it — otherwise it would stop demonstrating the bypass. It also pins that a genuine top-level alternation stays accepted and stays anchored at both ends, since that is what the group is for.

While in there, an adversarial review pass found a second one of the same shape: `[:` opens a POSIX class in RE2 only when a `:]` follows, so `[[:/-~]` is an ordinary class containing `/`. `endOfBracketExpression` returned -1 for it, which made the guard abandon the whole pattern, and `https://[[:/-~]+\.example\.com` matched `https://e/v/i/l/x.example.com`. Also fixed, also tested.

Both are tokenization divergences rather than missing cases, which is worth noting: per-atom analysis is only sound while the scanner tokenizes the way RE2 does. An earlier pass had already made an unanalyzable atom reject the pattern instead of being skipped, but that does not help here, because nothing failed to compile — the atom was never extracted.

pkg/octosts/org_trusted_issuers_store.go
salekseev: You're right, and this is a gap I shipped knowingly — so let me be straight about why, and about which of your two suggested fixes actually works.

**It is implemented, tested, and deliberately not in this PR.** `confirmNoAccess` re-enumerates via a new `ghinstall.Manager.GetAllFresh` before concluding absence, and reads any installation the first pass could not see: if one can read `.github`, enforcement turns on in the same request; if they all agree, the conclusion is earned. A failed confirmation is ignorance and falls through to stale/fail-closed. It lives on `feat/octosts-confirm-no-access` and is tracked in #1546.

The reason it isn't here: `GetAllFresh` is in #1543, so including the caller would make this PR unmergeable until that one lands. You'd asked for this work to be split, so I split at the only seam that made the two PRs independent — and this gap is exactly what that cost. Your review says the cost is too high, which I think is the right call.

**On your two suggested fixes — the first works, the second doesn't.**

"Bypass the negative cache for this enumeration" is `GetAllFresh`, i.e. #1543 + the confirm.

"Treat an app missing from the result as unknown/fail-closed" I looked at hard, because it would fix this PR standalone and preserve the independence. It doesn't work: from `GetAll`'s result you cannot distinguish *not installed* from *hidden by the negative cache*. `manager.GetAll` maps both to `(nil, nil)`. The only signal available in-process is `s.appCount` (`len(managers)`), and an App legitimately not installed on an org is the **normal** case for a multi-app deployment — so `len(installs) < appCount` would fail closed for most orgs on most lookups. That indistinguishability is precisely why `GetAllFresh` exists.

So closing it here means this PR depends on #1543. Your call on sequencing, and either is fine by me:

1. **Fold the confirm into this PR.** Fixes your finding here; this PR then requires #1543 to merge first. One fewer PR, one dependency.
2. **Keep as-is, land the follow-up right after.** Two independent PRs now, gap open for the window between merges.

Worth flagging regardless: there is already a hard ordering constraint between this PR and #1543 that has nothing to do with the confirm. #1543 adds `GetAllFresh` to the `Manager` interface, and this PR adds two fakes implementing `Manager` in a file that doesn't exist on `main`, so #1543 can't stub them. Whichever merges second leaves `main` with a broken test build — `go build` passes, there's no textual conflict, and GitHub shows both MERGEABLE, so the first merge is green and the damage shows on the second. Details and the fix are in both descriptions. Given that, option 1 may be the simpler thing to review.

Say which you'd prefer and I'll push it.

salekseev:
> @salekseev thanks. Taking a look. 4K lines added is a bit brutal but 60% of it is tests. We do need to drop the `.vscode/settings.json` though. I am going to have to take you up on your offer to split this. It is a great PR but splitting it up will be necessary IMO. You good with that?

@jmeridth also atill waiting on your call re: `.vscode/settings.json` - the question is in https://github.com/octo-sts/app/pull/901's description under "Open question". Short version: main's existing glob is /.github/chainguard/*, which matches everything, so once the allowlist file exists an editor validates it against the TrustPolicy schema and reports errors on a valid file. Four options listed there; happy with any of them, including removing the file from the repo entirely.

I can't seem to be able to get this change to meaningfully under 4K LOC given all the test fixtures and json schemas.

pkg/octosts/org_trusted_issuers_store.go
jmeridth: Let's go with **Option A** -- fold both `GetAllFresh` and `confirmNoAccess` into this PR so the control ships with its own enforcement intact. I'd rather not merge the allowlist in a bypassable state, even briefly, for a control whose whole job is stopping repo-policy writers from reaching arbitrary issuers. Folding both in also sidesteps the #901<->#1543 interface/fakes build-break you flagged, and #1543 can close.

And you're right that my second suggestion doesn't work -- in a multi-app deployment an App legitimately not installed on an org is the normal case, so `len(installs) < appCount` would fail closed for nearly every lookup. A fresh re-enumeration (`GetAllFresh`) is the only sound fix; thanks for the clear write-up.

One rebase heads-up: #1544 just merged, which changed the `octosts.go` event-emission block (now `if s.metrics && s.ceclient != nil`) that this PR also touches -- expect a small conflict there to reconcile. Once it's folded in and rebased, we'll do a full re-review of the combined PR.

jmeridth:
On `.vscode/settings.json`: let's drop it from this PR. Revert it to `main` and don't touch it here. One correction on the reasoning, though: the README doesn't actually point at `settings.json`. Since #1270 it recommends the inline `# yaml-language-server: $schema=...` header (README:75-76) and the examples use it directly, so there's no README dependency on the settings file to preserve. For the allowlist, just add that same inline header to the example/testdata pointing at `octosts.OrgTrustedIssuers.json`, the way the trust-policy examples already do it. I'll open a separate issue to remove `.vscode/settings.json` from the repo entirely, since it's now redundant with the inline-header approach and is editor-specific. That keeps this PR focused on the feature.

salekseev:
Done — `.vscode/settings.json` is reverted to `main` and out of this PR's diff entirely (23 files → 22).

Thanks for the correction, and you're right: I had claimed the README points at the settings file, and it doesn't. Since #1270 it recommends the inline `# yaml-language-server: $schema=...` header and the examples use it directly, so there was never a dependency on the committed editor config to keep in step. My "the README points at it" framing in the description was wrong and is now fixed there too.

The allowlist gets that same inline header instead: on the README example, and on all four testdata fixtures pointing at `octosts.OrgTrustedIssuers.json`. While adding those I noticed my fixtures were also missing the copyright headers every other testdata YAML in the repo carries, so they have those now.

One small thing I'll note rather than quietly leave: no testdata in the repo used the inline header before this, so the fixtures are now slightly ahead of the existing `*.sts.yaml` ones on that convention. Harmless, and it means opening a fixture gives you completion — but if you'd rather fixtures stay header-free for consistency, say so and I'll drop it from them and keep it on the README example only.

And for the record on the consequence of reverting: `main`'s glob is `/.github/chainguard/*`, so anyone still using the committed settings file will see an editor validate `trusted-token-issuers.yaml` against the **TrustPolicy** schema and flag errors on a valid file. The inline header takes precedence in vscode-yaml, so the documented path avoids it, and your separate removal issue closes it properly. Not fixing it here.

jmeridth:
:robot: (summary) AI-assisted re-review, 4 models (Opus 4.8, Sonnet 5, Haiku 4.5, Fable 5).

Both prior blocking findings are fixed and verified:

**HIGH (multi-app `GetAll` truncation): verified closed.** Reproduced the original scenario end-to-end with real `ghinstall.NewRoundRobin` (the `enumMgr` double is what hid it before): `GetAll` returns only the blind App, `GetAllFresh` returns both, the old path would have cached allow-all, but `orgIssuerLookup` now returns `Present` and denies a non-allowlisted issuer. `confirmNoAccess` calls `GetAllFresh` to re-enumerate past the negative cache and fails closed on any partial/failed fresh walk. `-race` plus a 64-goroutine stress run showed no races and no permanent misses.

**MEDIUM (unbalanced-paren anchor bypass): still fixed.** And nice catch on your two self-found fail-opens (the spam-`422`-as-agreement path and the stale-reseed TTL extension); both verified sound, including against `golang-lru` `Get` not renewing the TTL.

**One blocking item, please fix in this PR (not just document):** `checkNoSlashMatchingAtom` is bypassable and its `KNOWN LIMITATION` note understates it. A `/`-matching element inside any group escapes the per-atom scan, not just a quantified one. I reproduced `?`, `{n,m}`, and alternation: `https://(x\.com/)?trusted\.example\.com` and `https://([a-z0-9.-]+/){1,3}trusted\.example\.com` both validate (webhook reports Valid) yet match `https://evil.com/x/trusted.example.com`, an attacker host with the trusted name demoted to a path segment. `([a-z0-9.-]+/)*trusted\.example\.com` is exactly what an owner would write for "subpaths of trusted.example.com", so the false-Valid is real. Close it rather than widen the disclosure: parse with `regexp/syntax.Parse` and reject the pattern if a `/`-matching element (a literal `/`, or a char class whose ranges include 0x2F) appears inside any repetition/optional/alternation node (`Star`/`Plus`/`Quest`/`Repeat`/`Alternate`). A fixed literal `/` at a fixed position stays allowed, so legit multi-segment issuer paths (e.g. `.../id/[A-Z0-9]+`) keep working. Details anchored inline.

Two non-blocking observations: (1) a persistently-inconclusive `confirmNoAccess` re-walks `ListInstallations` per exchange with no backoff during a GitHub outage (fails closed; healthy case bounded to the 5-min cache), worth a conscious accept. (2) Coverage nits: no "422 + rate-limit headers" row in `TestClassifyMintError`, and no true-concurrent goroutine `-race` test for `manager`; the code is correct, these are durability guards.

Huge progress overall. Once the regex guard is closed in-PR, this is good to go.

pkg/octosts/org_trusted_issuers.go
```diff
@@ -0,0 +1,315 @@
...
+// atom lets a pattern span past the host, so "https://.*\.example\.com" would also permit
+// "https://totally-evil.com/x.example.com". BEST-EFFORT FOOTGUN GUARD, NOT A SECURITY
+// BOUNDARY — the author is the org owner.
+//
+// The test is EMPIRICAL — each atom compiled alone as "^atom$" and asked whether it matches
+// "/" — because banned spellings cannot be enumerated: "\S", "[^\n]" and "[[:ascii:]]" match
+// "/" without containing ".", and "[.-9]" holds "/" by range. A LITERAL "/" stays allowed;
+// unterminated brackets are left to regexp.Compile.
+//
+// KNOWN LIMITATION, structural rather than a missing case: being per-ATOM, a literal "/"
```
jmeridth: :robot: (medium) The per-atom guard is bypassable and this `KNOWN LIMITATION` understates it: a `/`-matching element inside any group escapes the scan, not only a quantified one. Reproduced with `?`, `{n,m}`, and alternation, all of which validate (webhook reports Valid) yet match `https://evil.com/x/trusted.example.com`:

- `https://(x\.com/)?trusted\.example\.com`
- `https://([a-z0-9.-]+/){1,3}trusted\.example\.com`
- `https://(good\.example\.com|evil\.com/x/trusted\.example\.com)`

Since the guard advertises catching patterns that read as safe, let's close this in the PR rather than document it. Parse with `regexp/syntax.Parse` and walk the tree: reject the pattern if a `/`-matching element (an `OpLiteral` containing `/`, or an `OpCharClass` whose ranges include 0x2F) appears anywhere inside a repetition/optional/alternation node (`OpStar`/`OpPlus`/`OpQuest`/`OpRepeat`/`OpAlternate`). Keep a fixed literal `/` at a fixed position allowed, so legit multi-segment issuer paths like `https://host\.example\.com/id/[A-Z0-9]+` still validate. That closes the whole class where the per-atom scan can't. Pin the three shapes above as regression tests.

salekseev: Fixed in `217ca2b` — closed rather than documented, as you asked. You were right that my `KNOWN LIMITATION` note understated it: I had framed it as quantified groups, and it is any variable-length or alternating construct.

I verified all three of your shapes against the code before changing anything, plus `+` and `*`, and confirmed each validated and matched a spanning issuer. One small correction on your first example: `https://(x\.com/)?trusted\.example\.com` does *not* match `https://evil.com/x/trusted.example.com` — the optional group is the literal `x.com/`, so the spanning issuer it actually admits is `https://x.com/trusted.example.com`. Same class, same bug, just a different demonstration string; the test row uses the one that matches so it can't pass vacuously.

Implemented your suggested approach: `checkNoSeparatorUnderQuantifier` parses with `regexp/syntax` and walks the tree, rejecting a `/`-matching element — an `OpLiteral` containing `/`, an `OpCharClass` whose ranges span 0x2F, or `.` — reachable under `OpStar`, `OpPlus`, `OpQuest`, `OpRepeat` or `OpAlternate`. Walking the AST rather than the text is what makes this cover the class: all six earlier bypasses were cases where my scanner tokenized differently than RE2 did, and this one sidesteps tokenization entirely.

A `/` at a fixed position stays legal, so the real multi-segment issuers still validate — the EKS and Keycloak patterns from the fixtures and README, and `https://login\.microsoftonline\.com/[a-f0-9-]+/v2\.0`. One nuance worth flagging since it wasn't in your description: a top-level alternation with **no** separator in it stays legal too (`https://a\.example\.com|https://b\.example\.com`), because the check keys on the `/` being under variable structure rather than on the structure alone. That keeps `TestAnchoringRejectsAlternationBypass` meaningful. An author who wants alternatives *with* paths writes separate `issuer_patterns` entries, which the error message tells them.

Tests: `TestCompileRejectsSeparatorUnderQuantifier` covers all five shapes — star, plus, optional, bounded repeat, alternation — each with a vacuity guard asserting the pattern really does match its spanning issuer, so a row fails loudly rather than silently if it stops demonstrating the bug. `TestCompileKeepsFixedPositionSlashes` pins the other half. The old limitation test told a future implementer to convert it to assert rejection rather than delete it, so that is what happened to it.

jmeridth (approved):

jmeridth:
@salekseev kudos to you. Thank you for your patience and hard work on this. You got it done. Thank you.

jmeridth:
Will work on getting this deployed asap.

PR #978, opened 2025-07-25
feat(webhooks): read file list from event payload

avidal:
The push event payload[0] already contains up to 3k files per commit and up to 2k commits.

Instead of using the Compare API and consuming rate limit just use the files from the event payload itself.

[0]: https://docs.github.com/en/webhooks/webhook-events-and-payloads#push


joedborg:
Hey @avidal, thanks so much for the PR!  Modified it a bit because we might need to support >20 commits (as rare as it hopefully is).  https://github.com/octo-sts/app/pull/1379

PR #988, opened 2025-07-30
webhook: Filter check_(suite|run) completed events

wlynch:
Only listen to created and requested event types.

- created: when the check is created by github for some work to do
- rerequested: when the user hits the retry button
- requested_action: (check run only) when the user hits another button on the check (corresponds to the `actions` field - technically not applicable for the webhook right now, but there's no harm in including it)

Mainly, this filters out completed events so we don't trigger a webhook check after we just finished updating a check, causing an infinite loop.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -12,6 +12,7 @@ import (
 	"mime"
 	"net/http"
 	"path/filepath"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


wlynch:
Moving to draft - there's more work needed here in order to try and better filter out the specific workflow events.

What's likely happening here is we're mixing up general repo checks with checks specifically for this app. But because we also encourage people to use this app to create additional checks, we need to be more careful about which checks we'll re-trigger on.

There's a subtle weakness here in that users of the octo-sts token can overwrite the checks that the webhook produces since they use the same identity. There isn't really additional exposure here though, since the webhook only tells you whether the policy is well-formed so you don't submit a busted config and is not intended to be a check for what policies are allowed - i.e. the failure mode is you might submit a broken config that then won't issue tokens. 🤷

ChrisJBurns:
@wlynch Thanks for this, in the case where the `check_run` events are disabled and only the PR events are sent, would this PR filter any events out in this case or is this PR only applicable in the case where `check_run ` events are also sent.

Hope that makes sense, it's been a while since I looked at this so I may be crossing some wires (apologies if thats the case).

jmeridth:
:robot: Closing as superseded. The infinite-loop this targets is now handled on `main` by the sender-based `isBotSender` filter added in #1404 — octo-sts's own check-run updates are bot-sent and skipped (`pkg/webhook/webhook.go`), so `completed` events no longer re-trigger a webhook check. Thanks for the original fix and write-up @wlynch; reopen if the action-level filtering turns out to still be wanted.

PR #1005, opened 2025-08-19
configure startup prober

cpanato:
```
│ Error: Error waiting for Updating Service: Error code 9, message: Revision 'octo-sts-00509-jsp' is not ready and cannot serve traffic. The user-provided container failed to start and listen on the port defined provided by the PORT=8080 environment variable within the allocated timeout. This can happen when the container port is misconfigured or if the timeout is too short. The health check timeout can be extended. Logs for this revision might contain more information.
```

manually updated the startup prober, and the deploy succeeded


cpanato:
i will close this, as that was a one off thing, if that happens again we reopen this PR

PR #1086, opened 2025-10-08
Add `/healthcheck` endpoints to facilitate deploying to k8s

jeffgran-dox:
Adds `/healthcheck` endpoints to the `app` and `webhook` servers so that k8s deployment can check when the node is up.

### DIFF
--- a/cmd/app/main.go
+++ b/cmd/app/main.go
@@ -20,6 +20,8 @@ import (
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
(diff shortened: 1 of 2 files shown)


mattmoor (commented):
Given that this is a GRPC service, I think I'd rather see us register the `healthpb` service and use a grpc probe, which thanksfully k8s now has.

I think it basically entails importing:
```
	"google.golang.org/grpc/health"
	healthpb "google.golang.org/grpc/health/grpc_health_v1"
```

and then doing something like this to register the service:
```
	// Register health check service
	healthServer := health.NewServer()
	healthpb.RegisterHealthServer(d.Server, healthServer)
```

PR #1248, opened 2026-03-04
return in the headers the app and installation id

cpanato:
this will return the app and install ids when we request a token

then we can see in the client side which install/app is being used

will be good for the quota monitoring, pr to come to update that


jmeridth (approved):

joedborg (approved):

jmeridth:
https://github.com/octo-sts/app/pull/1309 covers getting the data back to GCP for quota monitoring. This PR is still good IMO. wdyt @cpanato?

joedborg:
I think this is also covered in https://github.com/octo-sts/app/pull/1336

jmeridth:
@joedborg agreed. closing.

PR #1270, opened 2026-03-20
docs: Mention yaml-language-server in file directly

pavelzw:
This is easier for people as then they don't need to dig through their VSCode settings. It then also will work for other IDEs like Zed.


pavelzw:
@cpanato any thoughts on this?

jmeridth (approved):

PR #1283, opened 2026-03-27
feat: add ORG_POLICY_REPO environment variable

simpsonw:
This PR adds a new environment variable called `ORG_POLICY_REPO` that allows you to specify the repository in your GitHub organization that should house org-level trust policy files.  This is useful for folks who are self-hosting octo-sts and would prefer to keep their trust policy files  in a dedicated or private repository separate from `.github`.

### DIFF
--- a/pkg/envconfig/envconfig.go
+++ b/pkg/envconfig/envconfig.go
@@ -38,6 +38,7 @@ type EnvConfig struct {
 type EnvConfigApp struct {
 	Domain          string `envconfig:"STS_DOMAIN" required:"true"`
 	EventingIngress string `envconfig:"EVENT_INGRESS_URI" required:"false"`
+	OrgPolicyRepo   string `envconfig:"ORG_POLICY_REPO" required:"false" default:".github"`
 }
 
 type EnvConfigWebhook struct {
--- a/pkg/envconfig/envconfig.go
+++ b/pkg/envconfig/envconfig.go
@@ -45,7 +45,11 @@ type EnvConfig struct {
 type EnvConfigApp struct {
 	Domain          string `envconfig:"STS_DOMAIN" required:"true"`
 	EventingIngress string `envconfig:"EVENT_INGRESS_URI" required:"false"`
-	OrgPolicyRepo   string `envconfig:"ORG_POLICY_REPO" required:"false" default:".github"`
+	// OrgPolicyRepo is the repository name (without owner) that holds
+	// org-scoped trust policies. It applies to every organization this
+	// instance serves. Defaults to ".github". Must not be empty; setting
+	// ORG_POLICY_REPO="" is rejected at startup.
+	OrgPolicyRepo string `envconfig:"ORG_POLICY_REPO" required:"false" default:".github"`
 }
 
 type EnvConfigWebhook struct {
@@ -60,6 +64,10 @@ func AppConfig() (*EnvConfigApp, error) {
 	var err error
 	if err = envconfig.Process("", cfg); err != nil {
 		return nil, err
+	}
+
+	if cfg.OrgPolicyRepo == "" {
+		return nil, errors.New("ORG_POLICY_REPO must not be empty")
 	}
 
 	return cfg, err
--- a/pkg/envconfig/envconfig.go
+++ b/pkg/envconfig/envconfig.go
@@ -56,6 +56,10 @@ type EnvConfigWebhook struct {
 	WebhookSecret string `envconfig:"GITHUB_WEBHOOK_SECRET" required:"true"`
 	// If set, only process events from these organizations (comma separated).
 	OrganizationFilter string `envconfig:"GITHUB_WEBHOOK_ORGANIZATION_FILTER"`
+	// OrgPolicyRepo is the repository name (without owner) that holds org-scoped
+	// trust policies and the org trusted-issuer allowlist. Defaults to ".github".
+	// Must not be empty; setting ORG_POLICY_REPO="" is rejected at startup.
+	OrgPolicyRepo string `envconfig:"ORG_POLICY_REPO" required:"false" default:".github"`
 }
 
 func AppConfig() (*EnvConfigApp, error) {
@@ -79,6 +83,10 @@ func WebhookConfig() (*EnvConfigWebhook, error) {
 	var err error
 	if err = envconfig.Process("", cfg); err != nil {
 		return nil, err
+	}
+
+	if cfg.OrgPolicyRepo == "" {
+		return nil, errors.New("ORG_POLICY_REPO must not be empty")
 	}
 
 	return cfg, err
--- a/pkg/envconfig/envconfig.go
+++ b/pkg/envconfig/envconfig.go
@@ -45,6 +45,7 @@ type EnvConfig struct {
 type EnvConfigApp struct {
 	Domain          string `envconfig:"STS_DOMAIN" required:"true"`
 	EventingIngress string `envconfig:"EVENT_INGRESS_URI" required:"false"`
+	OrgPolicyRepo   string `envconfig:"ORG_POLICY_REPO" required:"false" default:".github"`
 }
 
 type EnvConfigWebhook struct {
--- a/pkg/octosts/octosts.go
+++ b/pkg/octosts/octosts.go
@@ -47,14 +47,15 @@ const (
 // NewSecurityTokenServiceServer creates an STS that exchanges OIDC tokens for
 // GitHub installation tokens. rrm handles installation selection; sticky (may
 // be nil) persists checks:write routing for check-run ownership.
-func NewSecurityTokenServiceServer(rrm ghinstall.Manager, sticky stickystore.Store, appCount int, ceclient cloudevents.Client, domain string, metrics bool) pboidc.SecurityTokenServiceServer {
+func NewSecurityTokenServiceServer(rrm ghinstall.Manager, sticky stickystore.Store, appCount int, ceclient cloudevents.Client, domain string, metrics bool, orgPolicyRepo string) pboidc.SecurityTokenServiceServer {
 	return &sts{
-		rrm:      rrm,
-		sticky:   sticky,
-		appCount: appCount,
-		ceclient: ceclient,
-		domain:   domain,
-		metrics:  metrics,
+		rrm:           rrm,
+		sticky:        sticky,
+		appCount:      appCount,
+		ceclient:      ceclient,
+		domain:        domain,
+		metrics:       metrics,
+		orgPolicyRepo: orgPolicyRepo,
 	}
 }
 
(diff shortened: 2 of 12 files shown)


joedborg:
Hey @simpsonw, thanks for the PR!  If you wouldn't mind fixing the conflicts, I'll re-review it.

simpsonw:
@joedborg Of course!  I've resolved the conflicts and added a couple of tests covering when the org policy repo is something other than `.github` as well.

jmeridth:
:robot: (low) AI-assisted review -- verified locally.

Trial-merged against current main: clean merge, builds, and the `octosts` + `envconfig` tests pass, including the new `TestExchangeCustomOrgPolicyRepo` (which uses a distinct permission so it'd catch a silent fallback to `.github`). Nice test.

Reads well: backward compatible (defaults to `.github`), and the trust model is unchanged since `ORG_POLICY_REPO` is operator config, not caller-controlled. LGTM.

Two minor, non-blocking notes: (1) the repo name is global per instance -- worth a doc line that it applies to every org the instance serves; (2) explicitly setting `ORG_POLICY_REPO=""` (vs leaving it unset) would break org-scoped lookups, since the default only applies when the var is absent.

simpsonw:
@jmeridth Great!  I've fixed the merge conflicts, added a comment about repo name affecting all orgs, and make it so we explicitly check for an empty ORG_POLICY_REPO at start up.  Let me know if there's other stuff I should address.

jmeridth:
:robot: (summary) AI-assisted re-review, 2 models (Opus 4.8 + Fable 5), against main with #901 merged. Build and tests pass, and #1283's own code is correct: the owner/repo resolution, the empty-`ORG_POLICY_REPO` startup rejection, and the backward-compatible `.github` default all check out. The issues are integration seams with code this PR doesn't touch, surfaced now that #901 (org-wide trusted issuers) has landed.

Heads up: my earlier LGTM predates #901 and didn't cover the webhook path, so this supersedes it.

1. (high) `ORG_POLICY_REPO` moves the org trust policies but not the #901 issuer allowlist, which stays hardcoded to `.github` (`pkg/octosts/org_trusted_issuers_store.go:262` and `:290`). An operator who relocates org policy to a private repo and puts `trusted-token-issuers.yaml` there, letting `org/.github` lapse, gets a silent fail-open: the read 404s, the allowlist is treated as absent, and all issuers are permitted. It's operator config rather than caller-controlled, and enforcement is opt-in, but it's an invisible loss of a fail-closed control under a plausible setup. Inline note on the wiring line.

2. (medium) The webhook can't validate org policies in a custom repo. The webhook binary reads `WebhookConfig()` (`EnvConfigWebhook`), which has no `OrgPolicyRepo` field, and `pkg/webhook/webhook.go:279/292/552` select org parsing with `EqualFold(repo, ".github")`. So a valid org trust policy in `my-policies` is strict-unmarshalled as a plain repo `TrustPolicy` and fails the check run, while the exchange path accepts it. The check-run verdict and runtime behavior diverge for exactly the deployments this PR targets, so this needs fixing regardless of how (1) is resolved. (Inline note on the `EnvConfigApp` field.)

3. (medium, docs) The var is undocumented: it's absent from the [environment variable table](https://github.com/octo-sts/app/blob/9878b045407e2817495c47aed065f3161173544c/README.md#L283), [README line 26](https://github.com/octo-sts/app/blob/9878b045407e2817495c47aed065f3161173544c/README.md#L26) ("Trust policies are checked into `.github/chainguard/...`") is now stale for non-default deployments, and the org-issuer section still says the allowlist [must live in `.github`](https://github.com/octo-sts/app/blob/9878b045407e2817495c47aed065f3161173544c/README.md#L222) (also [line 82](https://github.com/octo-sts/app/blob/9878b045407e2817495c47aed065f3161173544c/README.md#L82)) with no note that `ORG_POLICY_REPO` doesn't move it.

Minor: the in-repo policy path stays `.github/chainguard/<identity>.sts.yaml` even inside the custom repo (works, but surprising and undocumented); and there's no test for the explicit `org/<orgPolicyRepo>` scope branch.

The call to make: either (A) thread `ORG_POLICY_REPO` through the issuer store, the path constant, and the webhook binary so every org-policy-repo reference honors it, or (B) scope the var to trust policies only and document that the allowlist stays in `.github`. Either way, (2) needs a fix since the feature's own validation path is broken for custom repos. Happy to re-review once you pick a direction.

pkg/octosts/octosts.go
```diff
@@ -282,12 +284,15 @@ func (s *sts) lookupInstallAndTrustPolicy(ctx context.Context, scope, identity,
...
 		otp.Repositories = []string{repo}
 	}
 
-	if repo == ".github" {
+	// If the repo is the org policy repo, then parse with an org policy even
+	// if the repo was specified explicitly because we will reject the
+	// repositories field in policies otherwise.
+	if repo == s.orgPolicyRepo {
 		tp = otp
 	}
```
jmeridth, line 316: :robot: (high) This wires `s.orgPolicyRepo` into the trust-policy lookup, but the sibling org-policy-repo references weren't updated, so the feature is only half-applied:
- the #901 issuer allowlist still reads `.github` (`pkg/octosts/org_trusted_issuers_store.go:262` mint scope and `:290` `GetContents`), and
- the webhook validator still keys on `EqualFold(repo, ".github")` (`pkg/webhook/webhook.go:279/292/552`) and never receives `ORG_POLICY_REPO`.

Net: with `ORG_POLICY_REPO=my-policies`, an operator who moves the issuer allowlist into `my-policies` and lets `org/.github` lapse silently loses enforcement (read 404 leads to allowlist absent leads to all issuers permitted). Either thread `orgPolicyRepo` through the issuer store + path constant + webhook binary, or keep the allowlist anchored to `.github` by design and document it. See the summary for the full picture.

jmeridth:
@simpsonw sorry. #901 got merged today. It needed to. It's been ongoing for a few days and was a beast. Causes you some headache. Apologies.

pkg/envconfig/envconfig.go
```diff
@@ -45,6 +45,11 @@ type EnvConfig struct {
 type EnvConfigApp struct {
 	Domain          string `envconfig:"STS_DOMAIN" required:"true"`
 	EventingIngress string `envconfig:"EVENT_INGRESS_URI" required:"false"`
+	// OrgPolicyRepo is the repository name (without owner) that holds
+	// org-scoped trust policies. It applies to every organization this
+	// instance serves. Defaults to ".github". Must not be empty; setting
+	// ORG_POLICY_REPO="" is rejected at startup.
+	OrgPolicyRepo string `envconfig:"ORG_POLICY_REPO" required:"false" default:".github"`
```
jmeridth, line 52: :robot: (medium) `OrgPolicyRepo` is only on `EnvConfigApp`. The webhook binary (`cmd/webhook/main.go`) reads `WebhookConfig()`, i.e. `EnvConfigWebhook`, which has no such field, so it never learns about a custom repo. Combined with the hardcoded `strings.EqualFold(repo, ".github")` selectors at `pkg/webhook/webhook.go:279/292/552`, a valid org trust policy pushed to `my-policies` is strict-unmarshalled as a plain repo `TrustPolicy` and fails the check run, while the exchange path accepts it. To make the webhook honor the setting, add `OrgPolicyRepo` to `EnvConfigWebhook` (and its loader), pass it into the validator, and replace the `.github` selectors with it. Needs fixing regardless of how the allowlist question is resolved.

simpsonw:
> @simpsonw sorry. #901 got merged today. It needed to. It's been ongoing for a few days and was a beast. Causes you some headache. Apologies.

@jmeridth 

No worries at all! The rebase went smoothly and it actually worked out well.  The changes surfaced some gaps in my implementation that needed fixing anyway.  This branch should be ready for review again.

jmeridth:
@simpsonw I think you'll need to rekick this one after the recent GitHub Actions incident. Checks are moving yet.

UDPATE: Update Branch button came back after incident officially mitigated. Rekicked.

jmeridth:
:robot: (summary) Re-review after Option A (`42dc45ac8`), 2 models (Opus 4.8 + Fable 5). Both prior blockers are CLOSED, verified by code trace and by throwaway behavioral tests (custom-repo allowlist now enforced → `PermissionDenied`; a custom-repo org policy now validates in the webhook instead of failing the check run). #901's fail-safe machinery (`confirmNoAccess`/`GetAllFresh`/`classifyMintError`/`exhaustiveNoAccess`) is byte-identical, so nothing was weakened. Build and the octosts/webhook/envconfig tests pass.

Before merge I'd ask for one thing: the README. It's the last piece of the original finding and it carries a real operator hazard, not just a doc nit.

Blocking:
- **Docs + migration warning.** README still has no mention of `ORG_POLICY_REPO` (env table, and the org-issuer section still says the allowlist must live in `.github`). More importantly: if an operator sets `ORG_POLICY_REPO=my-policies` while their allowlist is still in `.github`, the read 404s, is treated as absent, and every issuer is permitted, a silent loss of enforcement (verified by test). The README needs to document the var and warn: create/move and protect the allowlist in the new repo before flipping the setting.

Non-blocking:
- **Two-binary config skew:** the app and webhook read `ORG_POLICY_REPO` from separate configs with nothing coupling them, so setting it on only one deployment makes the webhook validate a different repo than the exchange enforces. Worth a docs note.
- **Test coverage:** `TestExchangeCustomOrgPolicyRepo` is a good non-vacuous test, but it only covers the trust-policy read; the two paths this PR fixes (custom-repo allowlist enforcement, custom-repo webhook validation) aren't covered. Inline note.
- **Minor:** the exchange path uses the raw `s.orgPolicyRepo` while the store uses the `policyRepo()` helper; harmless today but defeats the helper's default. Inline note.
- **Nits:** a couple of log/comment strings still say `"%s/.github"` literally, misleading under a custom repo.

The in-repo path layout (allowlist at `<repo>/.github/chainguard/...`) is intentional and consistent across the store read, exchange read, webhook glob, and dir scan, so it's fine as long as the README explains it. Nice work closing both blockers cleanly.

pkg/octosts/octosts.go
```diff
@@ -293,12 +302,15 @@ func (s *sts) lookupInstallAndTrustPolicy(ctx context.Context, scope, identity,
 
 	owner, repo := path.Dir(scope), path.Base(scope)
 	if owner == "." {
-		owner, repo = repo, ".github"
+		owner, repo = repo, s.orgPolicyRepo
```
jmeridth: :robot: (low, non-blocking) This uses the raw `s.orgPolicyRepo`, but the store reads the allowlist via the `policyRepo()` helper (which defaults `""` → `.github`). With an empty field the exchange would resolve the org policy against repo `""` while the store defaults to `.github`. Not reachable through the binaries (envconfig rejects an empty value), so it's latent, but using `s.policyRepo()` here and at `:313` keeps the defensive default uniform.

pkg/octosts/octosts_test.go
```diff
@@ -257,6 +258,78 @@ func TestExchange(t *testing.T) {
 	}
 }
 
+// TestExchangeCustomOrgPolicyRepo verifies that an org-scoped exchange reads
+// its trust policy from the repo named by ORG_POLICY_REPO rather than the
+// hardcoded ".github" default.
+func TestExchangeCustomOrgPolicyRepo(t *testing.T) {
```
jmeridth, line 264: :robot: (low, non-blocking) Nice non-vacuous test (the `contents: read` vs `.github`'s `pull_requests: write` means a silent fallback would fail it), but it only covers the trust-policy read. The two behaviors this PR set out to fix aren't covered: (1) a populated allowlist enforced from a custom repo, and (2) the webhook validating an org policy in a custom repo. Both are cheap to add and would lock in the fixes against regression.

jmeridth:
@simpsonw the Blocking section of the big comment above is the piece to focus on. The rest is notes or nice to have. Thank you for your work on this and your patience with my feedback. 🙇

simpsonw:
@jmeridth Thanks for the detailed review!  I believe I've addressed everything, let me know if there are further concerns.

jmeridth (approved):

PR #1141, opened 2025-11-16
adds support for AWS KMS

ChrisJBurns:
Recreates https://github.com/octo-sts/app/pull/892

For some reason the underlying repo that had the original PR code was deleted. This PR restores that contribution because like others I would still very much be interested in hosting a version of the OctoSTS app on AWS using AWS KMS and Secrets Manager.

implements https://github.com/octo-sts/app/issues/438

PR #1317, opened 2026-04-14
feat: multi-org github app routing with yaml config

lee2sh:
Multi-Org GitHub App Routing

Problem

octo-sts supports multiple GitHub Apps, but only as a flat pool for rate-limit scaling — all apps serve all orgs indiscriminately. There's no way to:

- Assign dedicated apps to specific GitHub organizations
- Use different credential types per app (e.g., KMS for org-A, injected PEM for org-B)
- Get a clear error when a request arrives for an org with no configured apps

The env-var config (GITHUB_APP_IDS, KMS_KEYS) is structurally limited to a single credential type across all apps.

Solution

Introduce a YAML config file (pointed to by APP_CONFIG_FILE) that maps orgs to their dedicated app pools, each with per-app credential configuration:

```
orgs:
  - name: org-a
    apps:
      - app_id: 111
        kms_key: projects/.../cryptoKeyVersions/1
  - name: org-b
    apps:
      - app_id: 222
        private_key: "${INJECTED_PEM}"  # env var expanded at load time
```

Let me know what you guys think!

### DIFF
--- a/modules/app/main.tf
+++ b/modules/app/main.tf
@@ -60,6 +60,35 @@ locals {
   #     --algorithm rsa-sign-pkcs1-2048-sha256 \
   #     --target-key-file key.data
   kms_keys = [for app in var.github_apps : app.key_version > 0 ? "${google_kms_crypto_key.app-keys[tostring(app.app_id)].id}/cryptoKeyVersions/${app.key_version}" : ""]
+
+  # Whether multi-org routing is enabled (at least one app has org_name set).
+  multi_org_enabled = anytrue([for app in var.github_apps : app.org_name != ""])
+
+  # Group apps by org_name for YAML config generation. Org names are
+  # lowercased here so that mixed-case entries (e.g. "Octo-STS" vs
+  # "octo-sts") fail at `terraform plan` if duplicated, instead of
+  # crashing the server at startup when appconfig.Validate rejects the
+  # generated YAML.
+  org_names = distinct([for app in var.github_apps : lower(app.org_name) if app.org_name != ""])
+  apps_by_org = {
+    for org in local.org_names : org => [
+      for app in var.github_apps : {
+        app_id  = app.app_id
+        kms_key = app.key_version > 0 ? "${google_kms_crypto_key.app-keys[tostring(app.app_id)].id}/cryptoKeyVersions/${app.key_version}" : ""
+      } if lower(app.org_name) == org
+    ]
+  }
+
+  # YAML config for multi-org routing.
+  app_config_yaml = local.multi_org_enabled ? yamlencode({
+    orgs = [for org in local.org_names : {
+      name = org
+      apps = [for app in local.apps_by_org[org] : {
+        app_id  = app.app_id
+        kms_key = app.kms_key
+      } if app.kms_key != ""]
+    }]
+  }) : ""
 }
 
 // Create a dedicated GSA for the IAM datastore service.
@@ -107,7 +136,16 @@ module "this" {
     "sts" = {
       image = var.images.app
       ports = [{ container_port = 8080 }]
-      env = [
+      env = local.multi_org_enabled ? [
+        {
+          name  = "APP_CONFIG_FILE"
+          value = "/etc/octo-sts/config.yaml"
+        },
+        {
+          name  = "STS_DOMAIN",
+          value = var.domain,
+        }
+        ] : [
         {
           name  = "GITHUB_APP_IDS"
           value = join(",", [for app in var.github_apps : app.app_id])
@@ -141,10 +179,53 @@ module "this" {
         name  = "EVENT_INGRESS_URI"
         value = { for k, v in module.sts-emits-events : k => v.uri }
       }]
+      volume_mounts = local.multi_org_enabled ? [{
+        name       = "app-config"
+        mount_path = "/etc/octo-sts"
+      }] : []
     }
   }
 
+  volumes = local.multi_org_enabled ? [{
+    name = "app-config"
+    secret = {
+      secret = google_secret_manager_secret.app-config[0].secret_id
+      items = [{
+        version = "latest"
+        path    = "config.yaml"
+      }]
+    }
+  }] : []
+
   notification_channels = var.notification_channels
+}
+
+// Store the multi-org YAML config in Secret Manager (only when multi-org is enabled).
+resource "google_secret_manager_secret" "app-config" {
+  count = local.multi_org_enabled ? 1 : 0
+
+  project   = var.project_id
+  secret_id = "${var.name}-app-config"
+
+  replication {
+    auto {}
+  }
+}
+
+resource "google_secret_manager_secret_version" "app-config" {
+  count = local.multi_org_enabled ? 1 : 0
+
+  secret      = google_secret_manager_secret.app-config[0].id
+  secret_data = local.app_config_yaml
+}
+
+resource "google_secret_manager_secret_iam_member" "app-config-accessor" {
+  count = local.multi_org_enabled ? 1 : 0
+
+  project   = var.project_id
+  secret_id = google_secret_manager_secret.app-config[0].secret_id
+  role      = "roles/secretmanager.secretAccessor"
+  member    = "serviceAccount:${google_service_account.octo-sts.email}"
 }
 
 // Allow the STS service to call the sign method on the keys in the keyring.
--- a/modules/app/main.tf
+++ b/modules/app/main.tf
@@ -60,6 +60,62 @@ locals {
   #     --algorithm rsa-sign-pkcs1-2048-sha256 \
   #     --target-key-file key.data
   kms_keys = [for app in var.github_apps : app.key_version > 0 ? "${google_kms_crypto_key.app-keys[tostring(app.app_id)].id}/cryptoKeyVersions/${app.key_version}" : ""]
+
+  # Whether multi-org routing is enabled (at least one app has org_name set).
+  multi_org_enabled = anytrue([for app in var.github_apps : app.org_name != ""])
+
+  # Group apps by org_name for YAML config generation. Org names are
+  # lowercased here so that mixed-case entries (e.g. "Octo-STS" vs
+  # "octo-sts") fail at `terraform plan` if duplicated, instead of
+  # crashing the server at startup when appconfig.Validate rejects the
+  # generated YAML.
+  org_names = distinct([for app in var.github_apps : lower(app.org_name) if app.org_name != ""])
+  apps_by_org = {
+    for org in local.org_names : org => [
+      for app in var.github_apps : {
+        app_id  = app.app_id
+        kms_key = app.key_version > 0 ? "${google_kms_crypto_key.app-keys[tostring(app.app_id)].id}/cryptoKeyVersions/${app.key_version}" : ""
+      } if lower(app.org_name) == org
+    ]
+  }
+
+  # YAML config for multi-org routing.
+  app_config_yaml = local.multi_org_enabled ? yamlencode({
+    orgs = [for org in local.org_names : {
+      name = org
+      apps = [for app in local.apps_by_org[org] : {
+        app_id  = app.app_id
+        kms_key = app.kms_key
+      } if app.kms_key != ""]
+    }]
+  }) : ""
+
+  # Env vars shared by both config modes. The sticky-store settings apply
+  # regardless of how the app pools are configured — in multi-org mode the
+  # sticky store is what preserves check-run ownership across each org's
+  # multi-app pool.
+  common_env = [
+    {
+      name  = "STS_DOMAIN"
+      value = var.domain
+    },
+    {
+      name  = "OCTOSTS_STICKY_STORE"
+      value = var.sticky_store
+    },
+    {
+      name  = "OCTOSTS_STICKY_STORE_FIRESTORE_PROJECT"
+      value = var.sticky_store == "firestore" ? var.project_id : ""
+    },
+    {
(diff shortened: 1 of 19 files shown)


joedborg:
Hi @lee2sh, thanks for the PR.  I like this approach.  Just to confirm, if you have a setup like us, we would do something like
```yaml
orgs:
  - name: "*"
    apps:
      - app_id: 111
        kms_key: projects/.../cryptoKeyVersions/1
      - app_id: 112
        kms_key: projects/.../cryptoKeyVersions/2
      - app_id: 113
        kms_key: projects/.../cryptoKeyVersions/3
```
...etc?

lee2sh:
> Hi @lee2sh, thanks for the PR. I like this approach. Just to confirm, if you have a setup like us, we would do something like
> 
> ```yaml
> orgs:
>   - name: "*"
>     apps:
>       - app_id: 111
>         kms_key: projects/.../cryptoKeyVersions/1
>       - app_id: 112
>         kms_key: projects/.../cryptoKeyVersions/2
>       - app_id: 113
>         kms_key: projects/.../cryptoKeyVersions/3
> ```
> 
> ...etc?

hi @joedborg, thanks for taking a look! I really appreciate your feedback.

Yes, that config does exactly what you'd expect. The router stores `{"*": <pool with apps 111/112/113>}`, and it will fall back to that wildcard pool. Inside the pool, the 3 apps are load-balanced based on the current implementation. 

Let me know if you have any more questions!

jmeridth:
:robot: AI-assisted review.

Thanks @lee2sh -- nice approach. Wrapping the existing round-robin + quota managers into per-org `OrgPool`s with a `*` wildcard fallback means no-config deployments behave exactly as they do today, which is the right call.

It's conflicting with `main` now -- could you rebase? Heads-up that it reworks the `NewSecurityTokenServiceServer` signature and the `cmd/app` wiring, which a couple of other in-flight PRs also touch (#1283, #1141), so you may hit some follow-on conflicts depending on merge order.

Happy to do a full review once it's rebased. Thanks for the contribution!

lee2sh:
> 🤖 AI-assisted review.
> 
> Thanks @lee2sh -- nice approach. Wrapping the existing round-robin + quota managers into per-org `OrgPool`s with a `*` wildcard fallback means no-config deployments behave exactly as they do today, which is the right call.
> 
> It's conflicting with `main` now -- could you rebase? Heads-up that it reworks the `NewSecurityTokenServiceServer` signature and the `cmd/app` wiring, which a couple of other in-flight PRs also touch (#1283, #1141), so you may hit some follow-on conflicts depending on merge order.
> 
> Happy to do a full review once it's rebased. Thanks for the contribution!

Hello @jmeridth, thank you for your review! I'm currently travelling but will rebase as soon as I get decent internet.

lee2sh:
Harro @jmeridth, sorry it took a long time to get it rebased, but finally got it rebased with main. Let me know how it looks!

jmeridth:
:robot: (summary) Full 4-model review (Opus 4.8, Sonnet 5, Haiku 4.5, Fable 5) now that it's rebased — thanks for the patience on that.

This is a genuinely strong PR. The core is solid and safe: with no `APP_CONFIG_FILE` it behaves exactly as today (all env apps land in one `*` wildcard pool backed by the same round-robin/quota manager — the entire org-trusted-issuers test suite runs through that wrapper unchanged), so existing deployments are unaffected. We also verified the org-issuer fail-closed machinery is intact (the store change is just a per-owner pool lookup), org isolation holds (a request for org A can't mint via org B's app — `manager.Get` only returns installations actually on that owner), no data races under `-race`, credentials are never logged, and config validation is strict and fails closed at startup. Nice approach wrapping the managers into per-org pools with the `*` fallback.

The findings are almost all in the multi-org path (Terraform, tests, webhook), so they only affect operators who opt into multi-org:

**Worth addressing before merge**
- **(high)** The Terraform multi-org `env` branch drops the `OCTOSTS_STICKY_STORE*` vars, so sticky check-run routing is silently off in multi-org deployments — an org pool with more than one app loses stable check-run ownership and GitHub 403s cross-app updates, with no startup error (Firestore still gets provisioned, so it looks wired). Inline. The Go supports sticky here; only the tf omits it.
- **(medium)** The `APP_CONFIG_FILE` toggle — the conditional the whole feature hinges on — has no test in `envconfig_test.go`. Inline.

**Worth it before multi-org is documented as usable (fast-follow is fine)**
- **(medium)** Migration footguns in the tf: a mix of apps with/without `org_name` silently drops the ones without; an org whose only app is filtered emits `apps: []` and the app crash-loops on `Validate`. Worth `plan`-time checks. Inline.
- **(medium)** The webhook isn't org-routed — it still uses `AppIDs[0]`, so policy-validation check runs for an org served only by a dedicated app will fail the installation lookup.

**Nits**
- **(low)** `os.ExpandEnv` runs over every config string: a principal who can write the config (but not the keys) could reflect an env var into a startup panic log, and any value containing `$` is silently mangled. Inline.
- **(low)** `Validate` accepts `"*"` as a hand-written org name (undocumented wildcard) — worth rejecting in YAML or documenting.
- **(low)** The `APP_CONFIG_FILE` / `orgs:`/`apps:` schema and the both-styles precedence aren't documented yet.

HUMAN (😄): Happy to re-review once the sticky-store wiring and the envconfig test are in. Really nice work overall.

modules/app/main.tf
```diff
@@ -107,7 +136,16 @@ module "this" {
     "sts" = {
       image = var.images.app
       ports = [{ container_port = 8080 }]
-      env = [
+      env = local.multi_org_enabled ? [
```
jmeridth: :robot: (high) This multi-org branch sets only `APP_CONFIG_FILE` and `STS_DOMAIN`, dropping the `OCTOSTS_STICKY_STORE*` vars the legacy branch below sets. With those unset, `baseCfg.StickyStore == ""`, so `cmd/app/main.go` never builds the sticky store (`sticky` stays nil) and `checks:write` policies lose stable check-run ownership — for an org pool with more than one app, GitHub 403s cross-app check-run updates, silently, with no startup error. `firestore.tf` still provisions the database, so it looks wired. Could this branch carry the same `OCTOSTS_STICKY_STORE*` vars when `var.sticky_store` is set?

modules/app/main.tf
```diff
@@ -60,6 +60,35 @@ locals {
   #     --algorithm rsa-sign-pkcs1-2048-sha256 \
   #     --target-key-file key.data
   kms_keys = [for app in var.github_apps : app.key_version > 0 ? "${google_kms_crypto_key.app-keys[tostring(app.app_id)].id}/cryptoKeyVersions/${app.key_version}" : ""]
+
+  # Whether multi-org routing is enabled (at least one app has org_name set).
+  multi_org_enabled = anytrue([for app in var.github_apps : app.org_name != ""])
```
jmeridth, line 65: :robot: (medium) Two migration footguns worth a `plan`-time check: (1) as soon as any app sets `org_name`, `multi_org_enabled` flips true and apps *without* `org_name` are dropped from the generated YAML while the legacy env vars stop being emitted — those apps silently stop serving. (2) An app with `org_name` set but `key_version = 0` is filtered by `if app.kms_key != ""` (~line 89); if that's an org's only app, the config emits `apps: []`, which `Validate` rejects, and the app crash-loops. A `precondition`/`check` for both would make these plan-time errors instead of silent outages / crash-loops.

pkg/envconfig/envconfig.go
```diff
@@ -16,10 +16,16 @@ type EnvConfig struct {
 	Port                       int      `envconfig:"PORT" required:"true"`
 	KMSKeys                    []string `envconfig:"KMS_KEYS" required:"false"`
 	KMSProvider                string   `envconfig:"KMS_PROVIDER" required:"false" default:"gcp"`
-	AppIDs                     []int64  `envconfig:"GITHUB_APP_IDS" required:"true"`
+	AppIDs                     []int64  `envconfig:"GITHUB_APP_IDS" required:"false"`
 	AppSecretCertificateFile   string   `envconfig:"APP_SECRET_CERTIFICATE_FILE" required:"false"`
 	AppSecretCertificateEnvVar string   `envconfig:"APP_SECRET_CERTIFICATE_ENV_VAR" required:"false"`
 	Metrics                    bool     `envconfig:"METRICS" required:"false" default:"true"`
+
+	// AppConfigFile, when set, points to a YAML file describing per-org GitHub
```
jmeridth, line 24: :robot: (medium) This `APP_CONFIG_FILE` toggle is the conditional the whole feature hinges on (it flips `GITHUB_APP_IDS` from required to optional), but `envconfig_test.go` isn't touched by this PR — nothing proves that setting `APP_CONFIG_FILE` bypasses the `GITHUB_APP_IDS is required` error, nor that omitting both produces the new error. A couple of `TestBaseConfig` cases would cover it cheaply.

pkg/appconfig/appconfig.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	expandEnvInPlace(m)
+	return m, nil
+}
+
+func (p envExpandingParser) Marshal(m map[string]any) ([]byte, error) {
+	return p.inner.Marshal(m)
+}
+
+func expandEnvInPlace(v any) {
```
jmeridth, line 105: :robot: (low) `expandEnvInPlace` runs `os.ExpandEnv` over every string leaf. Two edges: (1) a principal who can write the config (but not the app keys) could reflect an arbitrary process env var into a startup panic message (e.g. `name: ${APP_SECRET_CERTIFICATE_ENV_VAR}` with zero apps panics with the expanded value) — an exfil path if a legacy env credential is ever co-set with `APP_CONFIG_FILE`; (2) despite the `${VAR}` comment, bare `$VAR` also expands and an unset var becomes `""`, silently mangling any value with a literal `$`. Fail direction is closed, so low.

jmeridth:
:robot: (summary) Re-reviewed the update — nicely done, the before-merge items are addressed and verified:

- **Sticky-store:** the `OCTOSTS_STICKY_STORE*` vars now live in a shared `common_env` that's `concat`'d onto both the multi-org and legacy branches, so multi-org deployments keep stable check-run ownership.
- **Toggle test:** `envconfig_test.go` now covers `APP_CONFIG_FILE` bypassing the `GITHUB_APP_IDS` requirement (and the inverse).
- **Bonus:** the two migration cases are now `terraform` `precondition`s that fail at plan time, and `os.ExpandEnv` is scoped to `${VAR}`-only. Build + `envconfig`/`appconfig` tests are green.

Two follow-ups, non-blocking (worth doing before multi-org is documented as usable): the webhook still routes via `AppIDs[0]`, so an org served only by a dedicated app would miss its installation during policy-validation check runs; and the `APP_CONFIG_FILE` / `orgs:` schema plus the `"*"` wildcard still want a short docs note. Both are fine as a fast-follow.

Resolving the addressed threads. Thanks for the quick turnaround!

jmeridth:
:robot: One more non-blocking thing worth landing with the feature: a short README section for the new config, since it's operator-facing (right now `APP_CONFIG_FILE` / the `orgs:` schema isn't documented anywhere). Not a merge blocker.

Here's a draft you can drop in and tweak — I pulled the field names straight from `pkg/appconfig` so it should be accurate:

````markdown
### Multi-org App routing (`APP_CONFIG_FILE`)

By default octo-sts uses a single flat pool of GitHub Apps (`GITHUB_APP_IDS` /
`KMS_KEYS`) that serves every org, for rate-limit scaling. To assign dedicated
Apps to specific orgs — e.g. to use different credentials per org — point
`APP_CONFIG_FILE` at a YAML file:

```yaml
orgs:
  # A dedicated pool for org-a (two Apps, GCP KMS credentials).
  - name: org-a
    apps:
      - app_id: 111
        kms_key: projects/p/locations/l/keyRings/r/cryptoKeys/k/cryptoKeyVersions/1
      - app_id: 112
        kms_key: projects/p/locations/l/keyRings/r/cryptoKeys/k/cryptoKeyVersions/2

  # A dedicated pool for org-b using an injected PEM instead of KMS.
  - name: org-b
    apps:
      - app_id: 222
        private_key: ${ORG_B_APP_PEM}   # ${VAR} is expanded from the environment

  # Optional catch-all: "*" serves any org not listed above.
  - name: "*"
    apps:
      - app_id: 999
        private_key_file: /etc/octo-sts/keys/fallback.pem
```

Each app sets **exactly one** credential: `kms_key` (a GCP KMS key-version
resource name), `private_key` (an inline PEM), or `private_key_file` (a path to a
PEM). String values support `${VAR}` expansion from the process environment.

Notes:
- When `APP_CONFIG_FILE` is set, the legacy `GITHUB_APP_IDS` / `KMS_KEYS` env vars
  are ignored.
- A request for an org with no matching pool and no `"*"` fallback is rejected
  (fail closed), not served by an arbitrary App.
- With `APP_CONFIG_FILE` unset, behavior is unchanged (a single flat pool).
- If you deploy with the bundled terraform module, set `org_name` on each entry
  in `github_apps` and the module generates this config for you (all-or-none
  `org_name`, and every app needs a KMS key).
````

Feel free to trim/reword — mainly want the schema, the credential options, the `"*"` fallback, and the "legacy vars ignored / unset == unchanged" notes in there.

jmeridth:
@lee2sh Please let me know if you want me to wait on the README idea or merge and I can do a quick follow PR for it.

jlee-nydig:
> @lee2sh Please let me know if you want me to wait on the README idea or merge and I can do a quick follow PR for it.

@jmeridth thank you for the review! I've updated the readme based on your suggestion.

jmeridth (approved):

PR #1318, opened 2026-04-15
fix: retry transient GitHub API errors in trust policy lookup

milesflo-bd:
## Problem

Intermittent `unable to find trust policy for "<identity>"` errors occur even when the policy file exists in the target repository.

`lookupTrustPolicy` fetches `.github/chainguard/<identity>.sts.yaml` from the GitHub API on every cache miss (the per-pod LRU expires after 5 minutes). The previous error handling had two gaps:

**1. Transport-level errors were not distinguished from file-not-found.**

When `atr.Token(ctx)` fails inside the HTTP transport — e.g. due to a transient 5xx or rate limiting during installation token creation — the error surfaces as an opaque transport error, not a `*github.ErrorResponse`. So `errors.As` never matched and all such errors fell through to the generic `codes.NotFound` / "unable to find trust policy" response, masking the real cause.

**2. No retry for transient errors.**

A single network blip or momentary GitHub API unavailability failed the request immediately with no recovery path.

## Fix

Wrap the `GetContents` call in `backoff.Retry` with exponential backoff (the same `cenkalti/backoff/v5` library already used in `provider.go`). Error classification:

- **404** → permanent, return `codes.NotFound` immediately (policy genuinely missing)
- **403 / 429** → permanent, return `codes.ResourceExhausted` immediately (rate limited)
- **Everything else** → transient, retry with exponential backoff

## Testing

- Existing `pkg/octosts` unit tests pass unchanged.

### DIFF
--- a/pkg/octosts/octosts.go
+++ b/pkg/octosts/octosts.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/bradleyfalzon/ghinstallation/v2"
+	"github.com/cenkalti/backoff/v5"
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	"github.com/coreos/go-oidc/v3/oidc"
 	"github.com/google/go-github/v84/github"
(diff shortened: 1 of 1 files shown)


pkg/octosts/octosts.go
```diff
@@ -393,29 +394,40 @@ func (s *sts) lookupTrustPolicy(ctx context.Context, base *ghinstallation.AppsTr
...
+						return nil, backoff.Permanent(status.Errorf(codes.ResourceExhausted, "GitHub API rate limit exceeded (429) for %q", trustPolicyKey.identity))
+					}
 				}
+				// Transient error (network blip, 5xx, token creation failure, etc.) — return
+				// a retriable sentinel. The original error is logged above; don't leak details.
+				return nil, status.Errorf(codes.Unavailable, "transient error fetching trust policy for %q", trustPolicyKey.identity)
 			}
-			return status.Errorf(codes.NotFound, "unable to find trust policy for %q", trustPolicyKey.identity)
+			return f, nil
+		}, backoff.WithBackOff(backoff.NewExponentialBackOff()))
```
jmeridth, line 422: `backoff.NewExponentialBackOff()` in v5 has no `MaxElapsedTime` field — it was moved to `WithMaxElapsedTime(d)` as a `RetryOption`. The 15-minute default lives inside `Retry` itself and is not overridden by `WithBackOff(...)`. On the hot token-issuance path during a GitHub degradation, a goroutine can hold for up to 15 min (or until the gRPC context deadline, which is implicit here).

The existing `maxRetry = 3` constant in this file suggests bounded retries are the house style.

```suggestion
		}, backoff.WithBackOff(backoff.NewExponentialBackOff()), backoff.WithMaxTries(3))
```

pkg/octosts/octosts.go
```diff
@@ -393,29 +394,40 @@ func (s *sts) lookupTrustPolicy(ctx context.Context, base *ghinstallation.AppsTr
...
+				// Permanent errors: don't retry.
+				var ghErr *github.ErrorResponse
+				if errors.As(err, &ghErr) && ghErr.Response != nil {
+					switch ghErr.Response.StatusCode {
+					case http.StatusNotFound:
+						return nil, backoff.Permanent(status.Errorf(codes.NotFound, "unable to find trust policy for %q", trustPolicyKey.identity))
+					case http.StatusForbidden:
+						return nil, backoff.Permanent(status.Errorf(codes.ResourceExhausted, "GitHub API rate limit exceeded (403) for %q", trustPolicyKey.identity))
+					case http.StatusTooManyRequests:
+						return nil, backoff.Permanent(status.Errorf(codes.ResourceExhausted, "GitHub API rate limit exceeded (429) for %q", trustPolicyKey.identity))
```
jmeridth, line 414: 429 (Too Many Requests) is GitHub's primary rate-limit response and is explicitly designed to be retried (GitHub includes a `Retry-After` header). Marking it `Permanent` means the backoff loop never retries it — the request immediately propagates `ResourceExhausted` to the outer app-rotation loop. If there's only one app configured, or all apps are simultaneously rate-limited, nothing retries.

Dropping this case lets 429 fall through to the transient path and get retried by the backoff loop you just added.

```suggestion
```

jmeridth:
@milesflo-bd Thank you for your contribution. This is a great idea. Just a couple suggestions 🙇

jmeridth:
Thanks again for this @milesflo-bd, and sorry for the slow turnaround. Closing because the code has drifted from `main` and the underlying path was rewritten since April.

The rate-limit half of what you identified has since shipped: `lookupTrustPolicy` was split into `fetchTrustPolicyRaw` + `lookupTrustPolicyWithRetry`, which on a 403/429 now serves a stale cached policy and rotates to the next app in the multi-app fleet. Layering an additional same-app `backoff.Retry` on top of that would stack two retry mechanisms, so this needs a fresh pass against the current structure rather than a rebase.

The other gap you called out is still real: transient non-rate-limit errors (5xx, network blips, token-creation failures) are still reported as `NotFound`. I opened #1589 to track that, credited to your original report here. Would be glad to have you take it on there if you are interested.

PR #1345, opened 2026-04-28
feat(ghinstall): capacity-aware fairshare routing across installations

nskaggs:
## Summary

Adds capacity-aware fairshare selection to the `roundRobin` manager. When quota data is available, requests route via three-tier waterfall — `comfortable` (`remaining ≥ SoftFloor`) → `tight` (`HardFloor ≤ remaining < SoftFloor`) → `last_resort` (`remaining < HardFloor`) — picking by `argmax(remaining)` within the highest non-empty pool.

`multiManager` is intentionally **not** changed: its consistent (scope, identity) hashing is the contract that preserves GitHub check-run ownership across token rotations, and any non-deterministic re-routing would cause 403s on check-run updates after a token refresh lands on a different App.

Quota state is sourced from `X-RateLimit-Remaining` / `X-RateLimit-Limit` headers via a transport-layer tap. Cold start (no quota data) falls back to the existing atomic-counter behavior.

## Why

When configured Apps have different installation rate-limit caps, the round-robin path treats them as equal slots — so high-cap installations can sit nearly idle while low-cap installations exhaust. `argmax(remaining)` (absolute, not ratio) routes heavy callers to the installation with the most absolute headroom, naturally absorbing more traffic on high-cap installations until they near parity with the rest.

## Tests

- Quota store: fresh / stale / overwrite / invalid input.
- Picker: nil-config / nil-store / cold-start / mixed-known / argmax / tier waterfall / not-installed skip / soft-floor demotion / tie-break alternation.
- `roundRobin`: integration with quota config, cold-start fallback (atomic counter spreads).
- Transport tap: header parsing, missing-context safety.
- Env config: floor validation.

### DIFF
--- a/cmd/app/main.go
+++ b/cmd/app/main.go
@@ -57,6 +57,18 @@ func main() {
 		}
 	}
 
+	// Capacity-aware routing: a shared QuotaStore is populated by the
+	// transport tap (X-RateLimit-Remaining headers on every GitHub response)
+	// and read by the manager selection logic. Cold start (no quota data
+	// yet) safely falls back to the existing FNV-hash / atomic-counter
+	// strategies.
+	quotaStore := ghinstall.NewQuotaStore(baseCfg.QuotaStaleAfter)
+	quotaCfg := &ghinstall.QuotaConfig{
+		Store:     quotaStore,
+		SoftFloor: baseCfg.QuotaFloorSoft,
+		HardFloor: baseCfg.QuotaFloorHard,
+	}
+
 	managers := make([]ghinstall.Manager, 0, len(baseCfg.AppIDs))
 	for i, appID := range baseCfg.AppIDs {
 		var kmsKey string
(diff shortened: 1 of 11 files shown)


joedborg:
Looks awesome, thanks!  One quick question.

Do we really want this acting on requests handed to `multiManager`?  `roundRobin` makes total sense (basically in order to counteract both the imbalance of max quota per install but also the imbalance of necessary `multiManager` hash based routing.  My worry here is that 

> Stickiness for check-run ownership comes for free from the caller-side oauth2.ReuseTokenSource 55-min token TTL.

isn't enough TTL and we regress (and I don't think it is because this was present when we had to implement `multiManager`).  Basically, we'll lose the deterministicness of `multiManager`, which handles the edge case of check runs needing to be created and updated by the same user (install in our case), if not the check will just sit in pending indefinitely.

joedborg (approved):
Thanks for addressing the `multiManager` feedback.  Looks good to me!

PR #1348, opened 2026-04-29
Replace multiManager with a persistent cache

joedborg:
I've finally bitten the bullet and added a persistent cache.  This only affects instances of OctoSTS that have multiple GitHub application installs (GHAI).

Currently, we round robin across the list of GHAIs per request, with some weighting based on the base quota of each GHAI.  This would lead to pretty much perfect distribution across the pool of GHAIs, except GitHub checks can only be updated by the user which created them.  In OctoSTS's case, this means needing to assign the same GHAI to the initial check create as well as any subsequent token requests, for example to update the check's status.  We've tried to get around this issue by, when the policy allows `check: write`, assigning the GHAI using a hash of the request's scope and identity to make the selection deterministic.  It works, but it's resulted in a very uneven distribution of GHAI quota, leading to quota exhaustion of one GHAI whilst others have barely been utilised.

I think the only way to have both even distribution across GHAIs and handle the checks edge case is to always assign using round robin and, if `checks: write`, cache the vended GHAI with the scope and identity hash as the key.  In order to do this we need a persistent cache, otherwise every redeploy would risk assigning a different GHAI and breaking checks updates.

To make this cache somewhat generic, I've added the `StickyStore` layer.  This means that, at a later stage, we (or anyone else) can add other key/value store backends.  There's also an in memory implementation if the redeploy issue isn't big enough of a worry.

For Chainguard specifically, I've opted for Firestore as we don't need lightspeed transactions but do want persistence.  There's a 30day TTL on the entries, bumped each `Get()`, so we don't fill up.  I think that's more than enough, if a check is being updated that was started 30days ago then there are probably other, much worse, issues.


cpanato (approved):
this is nice

lgtm

PR #1370, opened 2026-05-08
Cache not installed in org

joedborg:
Similar to https://github.com/octo-sts/app/pull/1334

We've seen a spike where many requests where made to an org without Octo-STS installed.  We should cache these misses to avoid quota exhausting churn.

### DIFF
--- a/pkg/ghinstall/ghinstall.go
+++ b/pkg/ghinstall/ghinstall.go
@@ -8,11 +8,13 @@ import (
 	"fmt"
 	"net/http"
 	"sync/atomic"
+	"time"
 
 	"github.com/bradleyfalzon/ghinstallation/v2"
 	"github.com/chainguard-dev/clog"
 	"github.com/google/go-github/v84/github"
 	lru "github.com/hashicorp/golang-lru/v2"
+	expirablelru "github.com/hashicorp/golang-lru/v2/expirable"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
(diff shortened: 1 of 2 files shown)


pkg/ghinstall/ghinstall.go
```diff
@@ -50,12 +50,21 @@ func New(atr *ghinstallation.AppsTransport) (Manager, error) {
 	}, nil
 }
 
+// negativeCacheConst is the sentinel value stored in the installation
+// LRU cache when a GitHub App is not installed for an owner. Valid
+// installation IDs are always positive, so 0 is unambiguous.
+const negativeCacheConst int64 = 0
```
jmeridth: Sentinel `0` collides with `(*Installation).GetID()` on a nil receiver/field — `GetID()` returns `0` when `i == nil || i.ID == nil`. If the API ever hands us an installation entry with a nil `ID` (corrupt/partial response, future schema drift), the positive-cache write a few lines down stores `0`, and the read at line 64 then misreads it as a negative-cache hit and returns `NotFound` for an org that **is** installed. The wire format is fine today, but the accessor's runtime behavior makes this a footgun.

Cheap fix: either guard the positive write with `if installID > 0` before `m.cache.Add(cacheKey, installID)` at line 90, or pick a sentinel `GetID()` can't produce (e.g. `-1`).

pkg/ghinstall/ghinstall.go
```diff
@@ -50,12 +50,21 @@ func New(atr *ghinstallation.AppsTransport) (Manager, error) {
...
+
 // Get returns the AppsTransport and installation ID for the given owner.
 // scope and identity are unused by the single-app manager; routing across
 // apps is handled by the roundRobin manager.
 func (m *manager) Get(ctx context.Context, owner, _, _ string) (*ghinstallation.AppsTransport, int64, error) {
 	cacheKey := fmt.Sprintf("%d/%s", m.atr.AppID(), owner)
 	if v, ok := m.cache.Get(cacheKey); ok {
+		if v == negativeCacheConst {
+			clog.InfoContextf(ctx, "negative install cache hit for %s", cacheKey)
```
jmeridth: This fires on every request in the exact spike scenario the PR is absorbing — we trade an API call for an Info log per request, and on the round-robin path we get one per manager per request. Drop to Debug:

```suggestion
			clog.DebugContextf(ctx, "negative install cache hit for %s", cacheKey)
```

pkg/ghinstall/ghinstall.go
```diff
@@ -84,6 +93,7 @@ func (m *manager) Get(ctx context.Context, owner, _, _ string) (*ghinstallation.
 		}
 		page = resp.NextPage
 	}
+	m.cache.Add(cacheKey, negativeCacheConst)
```
jmeridth: No TTL and no invalidation hook on this entry — bigger concern than it looks.

The PR description points at #1334 as the model, but #1334 used `expirablelru.NewLRU` with a 5-minute TTL (see `pkg/octosts/octosts.go:61`). This cache is `lru.TwoQueueCache` with no expiration, and the webhook handler in `pkg/webhook/webhook.go` only subscribes to `pull_request` / `push` / `check_suite` / `check_run` — there's no `installation` event handler and no callsite that does `cache.Remove`.

Result: an org that installs the App **after** a negative entry lands in the cache will get persistent `NotFound` until 2Q evicts it (size-driven, non-deterministic at size=200) or the pod restarts. `GetByInstallation` inherits the same staleness, which would also break sticky-session resume on a re-install.

Two fixes that would close this:
- Switch to `expirablelru.NewLRU[string, int64](200, nil, 5*time.Minute)` to mirror the trust-policy cache.
- Or add an `Invalidate(owner)` method on `Manager` and wire an `installation` event handler in the webhook to call it on `created` / `new_permissions_accepted`.

jmeridth (commented):
Three correctness notes inline. Caching the miss is the right call — most of the comments are about hardening the edges.

pkg/ghinstall/ghinstall.go
joedborg: Good call, have made a separate cache.  I don't think we'd ever actually hit this, but worth having a real answer if we do.

pkg/ghinstall/ghinstall.go
joedborg: We have this as Info in other places, so keeping convention

pkg/ghinstall/ghinstall.go
joedborg: This is now redundant

jmeridth (approved):

PR #1372, opened 2026-05-11
chore(stickystore): use docstore for sticky store

lee2sh:
Summary:

swaps out firestore-only implementation with gocloud.dev/docstore.

config example: 
# In-memory (tests, local dev)
OCTOSTS_STICKY_STORE_URL="mem://sticky/key"

# Firestore (GCP) — Chainguard's prod
OCTOSTS_STICKY_STORE_URL="firestore://projects/my-project/databases/(default)/documents/sticky-routes?name_field=key"

# DynamoDB (AWS)
OCTOSTS_STICKY_STORE_URL="dynamodb://sticky-routes?partition_key=key&region=us-east-1"

I like leveraging docstore here but was wondering if we could cover multi-cloud (GCP, AWS) by leveraging gocloud.dev/docstore. It also supports in-memory docstore. 

However, this is going to be a breaking change in configs. Let me know what you guys think!


joedborg:
Thanks for the PR @lee2sh!  I think we're going to keep the current, generic, setup as it gives us the flexibility to add any kind of backend (e.g. Redis or etcd or even SQL).

lee2sh:
Hey @joedborg thanks for looking into the PR!

Just wanted to clarify one thing in case it's useful. The `stickystore.Store` interface is preserved so a future redis/sql backend would still be a direct store implementation similar as today. docstore implementation is wired in as one implementation of store, not a replacement for the interface, so the door to other backends stays open. 

with this being said, totally understand the preference to keep things simple and happy to close this out if it's not the direction you want to go. Just wanted to share/clarify. No strong attachment either way.

PR #1379, opened 2026-05-14
Improve webhook to make less GitHub API calls

joedborg:
Uses less GH API calls if there are less than 20 commits (1 page).


cpanato (approved):
sweet

Issue #1395, opened 2026-05-27
Feature Request: Support for Github Enterprise Server (GHES)


sonikro:
### Description

Octo STS is defaulting to the Public GitHub API, and does not allow us to change the GitHub Base URL if we have GitHub Enterprise Servers running (mycompany.github.com).

## Feature Request

Add a new config that allows us to deploy an Octo STS App that can communicate with Custom Github Enterprise Servers API Endpoints.

## Ideas

I think this could be as simple as adding a new environment variable, and then refactoring all points of the code that talk to GitHub to rely on a config instead of hardcoding it to api.github.com.

jmeridth:
:robot: Delivered by #1400. `GITHUB_BASE_URL` sets the GitHub API base URL and is wired through the app, `ghinstall`, the STS client, and the webhook via `github.WithEnterpriseURLs` (validated as an https URL at startup). Closing as implemented.

PR #1400, opened 2026-06-02
feat: add support for GitHub Enterprise Server with configurable API base URL

sonikro:
Solves https://github.com/octo-sts/app/issues/1395

This was tested in 2 deployments, one for github.com and another one for a private Github Enterprise Server.


jmeridth:
:robot: AI-assisted review.

Thanks @sonikro -- good to see GHES support, and nice call validating `GITHUB_BASE_URL` (https-only, operator-set), so there's no untrusted-input surface.

It's conflicting with `main` now -- could you rebase? (It reworks the `NewSecurityTokenServiceServer` signature and `cmd/app` wiring that a few other in-flight PRs also touch.)

One question for when you're back on it: this handles where octo-sts calls *out* to the GHES API, but a GHES Actions caller also presents an OIDC token from a **different issuer** (e.g. `https://github.example.com/...`), and the token-verification / JWKS-discovery path isn't touched here. Have you been able to test the full exchange against a real GHES token -- i.e. does discovery reach the GHES issuer -- or is that intentionally out of scope for this PR? Happy to do a full review once it's rebased.

sonikro:
> One question for when you're back on it: this handles where octo-sts calls _out_ to the GHES API, but a GHES Actions caller also presents an OIDC token from a **different issuer** (e.g. `https://github.example.com/...`), and the token-verification / JWKS-discovery path isn't touched here. Have you been able to test the full exchange against a real GHES token -- i.e. does discovery reach the GHES issuer -- or is that intentionally out of scope for this PR? Happy to do a full review once it's rebased.

Yes, the JWKS Discovery path works when using a issuer from the enterprise server, so nothing had to be changed on that side. 

I merged the changes from main and resolved the conflicts.

jmeridth:
:robot: Heads-up: #1141 just merged, reworking the signing/`ghtransport` wiring and `envconfig` that this PR also touches (it replaced `pkg/gcpkms` with a `pkg/kms` factory). You'll need a rebase onto the new structure -- the GHES base-URL change itself should still slot in fine; flagging so the conflict isn't a surprise. Still curious about the OIDC-issuer question from the earlier review whenever you pick it back up.

sonikro:
> 🤖 Heads-up: #1141 just merged, reworking the signing/`ghtransport` wiring and `envconfig` that this PR also touches (it replaced `pkg/gcpkms` with a `pkg/kms` factory). You'll need a rebase onto the new structure -- the GHES base-URL change itself should still slot in fine; flagging so the conflict isn't a surprise. Still curious about the OIDC-issuer question from the earlier review whenever you pick it back up.

Conflict resolved.

jmeridth (approved):

PR #1403, opened 2026-06-05
fix(webhook): skip full directory scan for new-branch check_suite events

blancadesal:
## What

When `handleCheckSuite` receives a `check_suite` event for a new branch (`BeforeSHA == zeroHash`) with no associated PRs, skip processing instead of listing the full `.github/chainguard/` directory and reading every policy file.

## Why

A new branch without associated PRs points at an already-validated commit — there are no policy file changes to validate. The current code reads all N policy files individually (N+1 API calls per event), which scales poorly on repos with many policy files and high branch creation rates, and provides no validation value.

## Notes

- New branches with associated PRs still go through the PR-file-diff path, which properly filters for `.sts.yaml` changes
- The non-branching path (`BeforeSHA != zeroHash`) is unchanged — it already filters via `CompareCommits`

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -362,6 +362,21 @@ func (e *Validator) handleCheckSuite(ctx context.Context, cs checkSuite) (*githu
 
 	var files []string
 	if cs.GetCheckSuite().GetBeforeSHA() == zeroHash {
+		// New non-default branch: skip if there are no associated PRs.
+		// A feature branch points at a commit already present in the
+		// repository and doesn't need a full directory scan. This
+		// avoids reading every policy file (O(N) API calls) on every
+		// new-branch event.
+		//
+		// We still process the default branch (initial commit) and
+		// any branch with associated PRs, since those may introduce
+		// new or modified policy files.
+		defaultBranch := cs.GetRepo().GetDefaultBranch()
+		headBranch := cs.GetCheckSuite().GetHeadBranch()
+		if headBranch != defaultBranch && len(cs.GetCheckSuite().PullRequests) == 0 {
+			log.Infof("skipping new non-default branch with no PRs")
+			return nil, nil
+		}
 		_, dirContents, _, err := client.Repositories.GetContents(ctx, owner, repo, ".github/chainguard", &github.RepositoryContentGetOptions{Ref: sha})
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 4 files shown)


pkg/webhook/webhook.go
```diff
@@ -362,12 +362,20 @@ func (e *Validator) handleCheckSuite(ctx context.Context, cs checkSuite) (*githu
...
+		//
+		// We still process the default branch (initial commit) and
+		// any branch with associated PRs, since those may introduce
+		// new or modified policy files.
+		defaultBranch := cs.GetRepo().GetDefaultBranch()
+		headBranch := cs.GetCheckSuite().GetHeadBranch()
+		if headBranch != defaultBranch && len(cs.GetCheckSuite().PullRequests) == 0 {
+			log.Infof("skipping new non-default branch with no PRs")
+			return nil, nil
 		}
```
jmeridth, line 379: The early-return guard looks right for skipping new non-default branches with no PRs. But the `GetContents` directory scan that was here before got removed entirely -- for cases that fall through (default branch initial commit, or non-default branch with PRs), `files` never gets populated from this block.

Concrete scenario: first push to `main` on a new repo that includes `.sts.yaml` files. `BeforeSHA == zeroHash`, `headBranch == defaultBranch`, no PRs. The guard doesn't fire, but `files` stays empty, the PR loop below adds nothing, and we silently return `nil, nil` with no check run created.

The comment on lines 371-373 says "We still process the default branch (initial commit)" but without the directory scan restored, that processing can't happen.

```suggestion
		// New non-default branch: skip if there are no associated PRs.
		// A feature branch points at a commit already present in the
		// repository and doesn't need a full directory scan. This
		// avoids reading every policy file (O(N) API calls) on every
		// new-branch event.
		//
		// We still process the default branch (initial commit) and
		// any branch with associated PRs, since those may introduce
		// new or modified policy files.
		defaultBranch := cs.GetRepo().GetDefaultBranch()
		headBranch := cs.GetCheckSuite().GetHeadBranch()
		if headBranch != defaultBranch && len(cs.GetCheckSuite().PullRequests) == 0 {
			log.Infof("skipping new non-default branch with no PRs")
			return nil, nil
		}
		_, dirContents, _, err := client.Repositories.GetContents(ctx, owner, repo, ".github/chainguard", &github.RepositoryContentGetOptions{Ref: sha})
		if err != nil {
			return nil, err
		}
		for _, file := range dirContents {
			files = append(files, file.GetPath())
		}
```

jmeridth (commented):
Worth noting: there's no test coverage for `handleCheckSuite` with `BeforeSHA == zeroHash` today. That's not something this PR needs to fix, but it would help catch this kind of regression. The existing tests all exercise the `push` event path.

pkg/webhook/webhook.go
blancadesal: yeah, that was meant to stay. thanks!

jmeridth (dismissed):

blancadesal:
added a few tests covering `handleCheckSuite`

jmeridth (approved):

PR #1412, opened 2026-06-12
Cache the installation client to stop minting a token per event

markusthoemmes:
Each webhook handler built a fresh github.Client via ghinstallation.NewFromAppsTransport on every event, and since that transport caches its installation token internally, throwing the client away after each event meant re-minting an access token on the first API call every single time — a large share of the webhook's REST traffic. Cache one client per installation ID instead: the transport then refreshes its token roughly hourly rather than on every event. The cache is a bounded LRU (size 200, matching the app's other caches) of sync.OnceValues constructors guarded by a mutex, so concurrent events for the same installation coalesce onto a single client and a single mint. Keying on installation ID alone is sufficient — GitHub installation IDs are globally unique and the webhook serves a single App.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -48,12 +48,13 @@ type Validator struct {
 
 	Organizations []string
 
-	// clients caches one client constructor per installation ID so the
-	// transport's token is reused (~1h) instead of re-minted per event. Keyed on
-	// installation ID alone: IDs are globally unique and the webhook serves one
-	// App. clientsMu guards the get-or-create; the OnceValues coalesces builds.
+	// clients caches one client per installation ID so the transport's token is
+	// reused (~1h) instead of re-minted per event. Keyed on installation ID
+	// alone: IDs are globally unique and the webhook serves one App. Building a
+	// client is cheap (the token is minted lazily on first use), so clientsMu
+	// can guard the whole get-or-create.
 	clientsMu sync.Mutex
-	clients   *lru.Cache[int64, func() (*github.Client, error)]
+	clients   *lru.Cache[int64, *github.Client]
 }
 
 // prActionsThatChangeFiles is the set of pull_request actions that can alter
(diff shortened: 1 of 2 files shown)


pkg/webhook/webhook.go
```diff
@@ -145,6 +158,37 @@ func (e *Validator) validatePayload(r *http.Request) ([]byte, error) {
...
+				Transport: ghinstallation.NewFromAppsTransport(e.Transport, installationID),
+			})
+			if e.Transport.BaseURL != "" {
+				return client.WithEnterpriseURLs(e.Transport.BaseURL, e.Transport.BaseURL)
+			}
+			return client, nil
+		})
+		e.clients.Add(installationID, build)
+	}
+	e.clientsMu.Unlock()
```
jml: I'd be mildly more comfortable if `getBuilder` was a separate method with a `defer c.clientsMu.Unlock()` and `clientForInstallation` called that and then called `build`

jml (dismissed):

pkg/webhook/webhook.go
```diff
@@ -45,8 +47,19 @@ type Validator struct {
 	WebhookSecret [][]byte
 
 	Organizations []string
+
+	// clients caches one client constructor per installation ID so the
+	// transport's token is reused (~1h) instead of re-minted per event. Keyed on
+	// installation ID alone: IDs are globally unique and the webhook serves one
+	// App. clientsMu guards the get-or-create; the OnceValues coalesces builds.
+	clientsMu sync.Mutex
+	clients   *lru.Cache[int64, func() (*github.Client, error)]
```
mattmoor: isn't the lru threadsafe?

mattmoor:
I'm not as familiar with the webhook as @wlynch

pkg/webhook/webhook.go
markusthoemmes: It is, but doesn't have a get-or-create op.

pkg/webhook/webhook.go
```diff
@@ -152,6 +165,37 @@ func (e *Validator) validatePayload(r *http.Request) ([]byte, error) {
...
+			if e.Transport.BaseURL != "" {
+				return client.WithEnterpriseURLs(e.Transport.BaseURL, e.Transport.BaseURL)
+			}
+			return client, nil
+		})
+		e.clients.Add(installationID, build)
+	}
+	e.clientsMu.Unlock()
+
+	// Build outside the lock so a slow mint doesn't serialize other lookups.
```
wlynch: I don't think you'll get much gain here - the transport doesn't fetch a token until request time. All this does is make a struct - https://github.com/bradleyfalzon/ghinstallation/blob/c8d44a371b850463d299cfd87e5c07d5e6d1c467/transport.go#L107-L117

I'd simplify this and make the LRU cache int64 -> *Client instead.

pkg/webhook/webhook.go
wlynch: idk if we need another method, but +1 on the defer.

wlynch (commented):
Looks good! Thanks for doing this.

pkg/webhook/webhook.go
markusthoemmes: you're right, I have no idea what I was thinking there. I've written too many caches in the near past 😂

wlynch (approved):

PR #1420, opened 2026-06-16
feat(observability): native OpenTelemetry logs, metrics, and trace export

MaxAnderson95:
_This PR was drafted by AI, but with heavy influence and direction from me. Particularly around ensuring this was entirely opt-in, and side effect free when not using it._

_If the code is crap, the feature isn't welcome, or an alternative implementation would be preferred, please feel free to let me know or to close this PR. We self-host octo sts ourselves and would simply love to be able to send telemetry to our OTLP backend._

## Summary

Adds opt-in, native OpenTelemetry export for **logs** and **metrics**, and closes a gap in **trace** export — all activated purely through canonical `OTEL_*` environment variables. Existing behavior is preserved when those variables are unset.

Today octo-sts can emit traces via `httpmetrics.SetupTracer` (from `terraform-infra-common`), but:

- there is no native OTLP export for **logs** or **metrics** (metrics are Prometheus-only via `/metrics`), and
- that tracer's OTLP path only forwards spans carrying a `gen_ai.*` attribute, so ordinary octo-sts spans are dropped when exporting to a generic OTLP backend.

This change lets an operator point octo-sts at any OTLP collector/backend with standard configuration, without changing the defaults that GCP/Cloud Run deployments rely on.

## What it does

- **Logs** — bridges `slog`/`clog` and the standard library `log` package to OTLP via the `otelslog` bridge, while preserving local stdout/stderr output. Each record is exported exactly once.
- **Metrics** — bridges the existing Prometheus default registry to OTLP using the contrib Prometheus bridge. No instrumentation is added or changed, and the `/metrics` endpoint is untouched.
- **Traces** — when `OTEL_TRACES_EXPORTER=otlp`, spans are exported directly over OTLP (bypassing the `gen_ai.*` span filter). When unset, the existing Google Cloud Trace default is preserved; `OTEL_TRACES_EXPORTER=gcp` continues to flow through the legacy helper.

## Activation & configuration

Everything is driven by canonical OpenTelemetry environment variables — there are no octo-sts-specific knobs:

- `OTEL_LOGS_EXPORTER`, `OTEL_METRICS_EXPORTER`, `OTEL_TRACES_EXPORTER` accept `otlp` or `none`; traces additionally accept `gcp`.
- OTLP endpoint/protocol come from the standard `OTEL_EXPORTER_OTLP_*` variables; both `http/protobuf` and `grpc` are supported.
- Resource identity uses `OTEL_SERVICE_NAME` / `OTEL_RESOURCE_ATTRIBUTES` plus the existing GCP/Cloud Run detection.

Behavioral guarantees:

- **Default-off:** with no `OTEL_*` exporter variables set, behavior is unchanged (Prometheus `/metrics`, stdout/stderr logs, GCP trace fallback).
- **Fail-fast:** an invalid explicit OTEL configuration fails startup rather than silently degrading.
- **Atomic setup:** if one signal fails to initialize, already-installed globals are rolled back.

## Testing

- Unit tests cover exporter/protocol parsing, partial-setup rollback, and that a single log call produces exactly one exported record (guarding against double export).
- Verified end-to-end against an OpenTelemetry Collector (`debug` exporter): logs, metrics (including the bridged Prometheus series), and traces all arrive over both OTLP `http/protobuf` and `grpc`; with no `OTEL_*` variables nothing is exported.

## Notes for reviewers

- During exploration, I noticed that telemetry is largely handeled by the external `terraform-infra-common` package (alongside `httpmetrics`). If you'd prefer this capability to live thererather than in this app, I'm happy to move it there. The only thing of note is that the [repo](https://github.com/chainguard-dev/terraform-infra-common) has PRs disabled currently.
- One intentional trade-off: when OTLP **log** export is enabled, the console `slog` format becomes the standard `slog` text format (`time=… level=… msg=…`). This avoids routing `slog` records back through the standard `log` package, which is what would otherwise double-export them. When OTLP logs are disabled, console output is unchanged.


jmeridth:
:robot: AI-assisted review.

Thanks for this, @MaxAnderson95 -- and for keeping it strictly opt-in and side-effect-free; that care shows.

Agreeing with your own first reviewer note: we'd prefer this not live in `octo-sts/app`. Telemetry here is centralized in `terraform-infra-common`, and I confirmed the gaps you're closing are real there today (no OTLP export for logs or metrics, and the tracer's OTLP path filters to `gen_ai.*` spans so ordinary infra spans are dropped) -- so your motivation makes sense. But the right home for the fix is tf-common, where it benefits every service, rather than a bespoke telemetry subsystem in this app.

I know tf-common currently has PRs disabled, which blocks the obvious path -- I'll raise that internally so there's a way forward rather than leaving you stuck. Appreciate the contribution and the clear write-up.

MaxAnderson95:
Hey no worries at all! I'll close this and open a new PR over there once able.

PR #1474, opened 2026-07-13
webhook: treat bare 403/429 ErrorResponse as rate-limited

phelian:
`isGitHubRateLimited` in `pkg/webhook` only matches the typed `*github.RateLimitError` / `*github.AbuseRateLimitError`. go-github returns those only when the response carries the primary (`X-RateLimit-Remaining: 0`) or secondary rate-limit markers. A rate-limited response without those markers surfaces as a plain `*github.ErrorResponse` and slips past the abort/CheckRun-skip protection added in #1404.

This adds a status-code fallback for `403`/`429` on `*github.ErrorResponse`, mirroring the generic handling already present in `pkg/octosts`'s `lookupTrustPolicy`. Aligns the two packages' rate-limit detection.

- `go test ./pkg/webhook/` passes; adds `TestIsGitHubRateLimited` covering typed errors, bare `ErrorResponse` 403/429, and non-rate-limit codes.
- `golangci-lint run ./pkg/webhook/`: 0 issues.


jmeridth:

Mirrors the existing `octosts.go` rate-limit detection (bare `*github.ErrorResponse` 403/429), which is the right call for consistency. Trial-merged clean against `main`; `pkg/webhook` tests pass, including the new `TestIsGitHubRateLimited` (typed errors, bare 403/429, 404 negative, nil) -- good coverage.

One intentional tradeoff worth naming: a bare `403` also covers genuine permission errors ("resource not accessible by integration"), so those will now be treated as rate-limited (back off / skip CheckRun) rather than surfaced -- but that matches what `octosts` already does, so it's consistent by design. LGTM.

**Merge-order note:** this pairs with #1475, which changes the rate-limited path to return `200` instead of `5xx`. Since #1474 only broadens *detection*, merging it **before** #1475 would make bare `403`/`429` (including genuine permission `403`s) return `500` -> GitHub treats the delivery as failed and redelivers -> amplifying load on the already rate-limited endpoint. Recommend merging **#1475 first** (or both together); the two touch different functions in `webhook.go`, so there's no conflict either way.

jmeridth (approved):

joedborg (approved):
Thank you @phelian !

PR #1475, opened 2026-07-13
webhook: acknowledge delivery when rate-limited instead of 5xx

phelian:
On a rate-limited push, `handleSHA` returned the rate-limit error, which `ServeHTTP` surfaces as a `500`. GitHub treats a non-2xx webhook delivery as failed and redelivers it, re-hitting the already rate-limited API and amplifying load.

This logs and returns `nil` so the delivery is acknowledged with a `200` while still skipping the CheckRun creation (the abort behaviour from #1404 is preserved; only the response code changes).

- `go test ./pkg/webhook/` passes; `TestWebhookPushAbortOnRateLimit` now also asserts a `200` response.
- `golangci-lint run ./pkg/webhook/`: 0 issues.


jmeridth:
:robot: (low) AI-assisted review -- verified locally.

This is the right fix: a `5xx` makes GitHub treat the delivery as failed and redeliver, re-hitting the rate-limited API and amplifying load. Returning `200` acknowledges the delivery while still skipping the CheckRun (the #1404 abort behavior is preserved). Trial-merged clean against `main`; `pkg/webhook` tests pass, and `TestWebhookPushAbortOnRateLimit` now asserts the `200`.

As noted on #1474: this should land **first** (or together with it). #1474 only broadens rate-limit *detection*; without this change, those newly-detected `403`/`429`s would return `500` and trigger redelivery. Merging #1475 first avoids that window. LGTM.

jmeridth (approved):

PR #1476, opened 2026-07-13
octosts: seed primary cache when serving a stale trust policy

phelian:
The stale-cache fallback added in #1404 serves a policy from the 1h `staleTrustPolicies` cache on a rate-limited fetch, but leaves the 5-minute primary `trustPolicies` cache empty. Every subsequent exchange during the rate-limit window therefore re-probes GitHub and falls back again.

This seeds the primary cache from the stale entry on a stale hit, so further exchanges resolve from cache without another round-trip. The stale value is already ≤1h old, so this only extends serving a known-good policy by at most the primary TTL.

- `go test ./pkg/octosts/` passes; `TestRateLimitServesStaleCache` now also asserts the primary cache is seeded.
- `golangci-lint run ./pkg/octosts/`: 0 issues.


jmeridth:

Sound optimization: seeding the 5-min primary cache from the stale entry avoids re-probing GitHub on every exchange during a rate-limit window. Correctly bounded -- once the 1h stale entry expires under a prolonged rate limit, serving stops (`ResourceExhausted`), so this can't extend serving beyond the existing window. Trial-merged clean against `main`; `pkg/octosts` tests pass, including `TestRateLimitServesStaleCache` asserting the primary cache is seeded.

One tradeoff worth naming: seeding the primary cache means a policy *tightening* served during a rate-limit window can lag up to the 5-min primary TTL **after** the rate limit clears -- exchanges keep hitting the seeded entry until it expires, rather than re-probing for the fresh policy immediately. It's consistent with the normal 5-min primary-cache freshness semantics, so a reasonable trade; just flagging it. LGTM.

jmeridth (approved):

PR #1511, opened 2026-07-28
feat(akv): add Azure Key Vault signing and secret provider

TomPeas:
### What 💡
Adds Azure Key Vault (AKV) as a third cloud backend alongside Google Cloud KMS and AWS KMS, for both:

- **JWT signing** — `pkg/kms/akv` implements the `kms.KMS` interface and registers as `KMS_PROVIDER=akv`.
- **Webhook secrets** — `pkg/secrets/akv` implements the `SecretProvider` interface for the same provider value.

**No new environment variables.** AKV reuses the existing `KMS_PROVIDER`, `KMS_KEYS` and `GITHUB_WEBHOOK_SECRET`, carrying fully-qualified Key Vault object identifiers the same way AWS uses ARNs and GCP uses `projects/*/cryptoKeys/*` paths:

```
KMS_PROVIDER=akv
KMS_KEYS=https://vault-a.vault.azure.net/keys/app-1,https://vault-b.vault.azure.net/keys/app-2/<version>
GITHUB_WEBHOOK_SECRET=https://vault-a.vault.azure.net/secrets/webhook-current,https://vault-a.vault.azure.net/secrets/webhook-previous
```

The version segment is optional; omitting it resolves to the current key/secret version. Managed HSM hosts (`*.managedhsm.azure.net`) work too.

### Why 🤔
Signing and webhook-secret storage were limited to GCP and AWS, forcing Azure-hosted deployments to take a dependency on another cloud for key management. This lets operators keep signing keys and webhook secrets inside their own Azure tenancy, using Managed Identity / Workload Identity for auth.

### How 📖
Follows the provider factory pattern introduced by the AWS KMS work (#1141), so AKV slots in without touching the transport layer.

- **`pkg/kms/akv`** — a `kms.KMS` (`ghinstallation.Signer` + `io.Closer`) backed by `azkeys`. Unlike GCP/AWS, Key Vault signs a **pre-computed digest**, so the signing string is SHA-256 hashed locally before requesting `RS256`. `Close()` is a no-op because `azkeys.Client` is REST-based and holds no closable resource. A `signerClient` interface seam allows unit testing without contacting Azure.
- **`pkg/secrets/akv`** — a `SecretProvider` backed by `azsecrets`, plus an exported `ParseSecretID` used to derive the vault URL from a secret identifier.
- **`pkg/internal/akverr`** — sanitizes Azure SDK errors, mirroring the existing `pkg/internal/awserr`. The Azure SDK renders the request URL (vault host, key name, version) and full response body into its error strings, and transport failures embed the host in the wrapped `net.DNSError`. `Sanitize` **fails closed**: unrecognised error types are replaced rather than passed through, so a new SDK error type cannot silently start leaking. Full errors still reach the protected structured log via `clog`.
- **`pkg/kms/kms.go` / `pkg/secrets/secrets.go`** — register `akv` in both factories. Because one `azsecrets.Client` is bound to one vault, `NewSecretProvider` validates each comma-separated `GITHUB_WEBHOOK_SECRET` entry individually (commas are legal in a URL path, so parsing the joined string would silently swallow all but the first) and rejects a list spread across multiple vaults.
- **`cmd/webhook`** — trims whitespace around comma-separated secret identifiers so validation and retrieval agree.
- Adds Azure SDK dependencies: `azidentity`, `azkeys`, `azsecrets`.

No changes to `pkg/ghtransport`, `cmd/app`, or the existing GCP/AWS providers.

### Test 🧪
`go build ./...`, `go test -race ./...` and `pre-commit run --all-files` (gofmt, goimports, go-mod-tidy, golangci-lint-full, misspell, woke, boilerplate) all pass.

33 new tests across three packages:

| Package | Tests | Coverage |
|---|---|---|
| `pkg/kms/akv` | 16 | 89.7% |
| `pkg/secrets/akv` | 8 | 100% |
| `pkg/internal/akverr` | 9 | 100% |

Notable coverage:

- **Signing correctness** — asserts the digest sent to Key Vault equals `sha256(signingString)` and is exactly `sha256.Size` bytes, that `RS256` is pinned, and that the signature is raw-URL base64. This pins the one behaviour that differs from GCP/AWS.
- **Identifier validation** — table-driven rejection of bare names, missing/`http` schemes, wrong object type, missing name, and foreign identifiers (AWS ARNs). These `recover()` explicitly to assert "error, never a panic": `azkeys.ID.Name()`/`azsecrets.ID.Name()` nil-dereference on identifiers with too few path segments, so the path guard is load-bearing.
- **Leak prevention** — asserts the vault host, key/secret name and response body never appear in returned errors, for API errors, auth failures, DNS/timeout/connection transport failures, and wrapped errors. Also asserts the sanitized error does **not** unwrap to the original `*azcore.ResponseError` via `errors.As`.
- **Real SDK exercise** — `pkg/secrets/akv` drives a real `azsecrets.Client` against a TLS `httptest` server with a fake `TokenCredential`, asserting the request path so a dropped version or wrong secret name is caught.
- **Multi-vault signing** — asserts the same key name in two different vaults resolves to independent clients.

### See 👀
Implements https://github.com/octo-sts/app/issues/1515
Builds on the provider factory from #1141 (AWS KMS support).

### Required checklist (before asking for a review)
- [x] I have performed a self-review of my code
- [x] Unit tests added/updated for new behaviour
- [x] Linting and formatting pass locally (`pre-commit run --all-files`)
- [ ] Documentation updated — no user-facing docs exist for signing backends (GCP and AWS are both undocumented, and `modules/` is GCP-specific). Happy to add an operator note covering the identifier format and the required Azure RBAC roles (`Key Vault Crypto User` for signing, `Key Vault Secrets User` for secret reads) if maintainers would like it in this PR.


pkg/azurekeyvault/azurekeyvault.go
```diff
@@ -0,0 +1,103 @@
...
+	client, err := azkeys.NewClient(vaultURL, cred, nil)
+
+	if err != nil {
+		return nil, fmt.Errorf("could not create client: %w", err)
+	}
+
+	return &akvSigner{
+		client: client,
+		key:    keyRef{name: keyName, version: keyVersion},
+		ctx:    ctx,
```
jmeridth: :robot: (medium) `akvSigner` stores this ctx and every `Sign` uses it. The ctx from `main` is `signal.NotifyContext(..., os.Interrupt)`, so after SIGINT all AKV signing fails with `context.Canceled` -- and the webhook keeps serving with signing permanently broken. `gcpkms` deliberately avoids this with `context.Background()`. Detach cancellation while keeping values:
```suggestion
		ctx:    context.WithoutCancel(ctx),
```

cmd/webhook/main.go
```diff
@@ -91,7 +106,26 @@ func main() {
 			}
 			webhookSecrets = append(webhookSecrets, resp.GetPayload().GetData())
 		}
-	} else {
+	case len(baseCfg.AKVUrls) > 0:
```
jmeridth: :robot: (medium) This gates the AKV webhook-secret path on `AKV_URLS`, but envconfig's mutual-exclusivity only counts `AKV_KEY_NAMES` -- so `AKV_URLS` can be set alongside cert/KMS signing. In that case a **literal** `WebhookSecret` gets split and treated as AKV secret *names*, and the fetch-error `log.Panicf("...secret %s...", name)` just below writes the literal secret value into logs. Suggest gating this branch on `len(baseCfg.AKVKeys) > 0` (matching the signing path) or rejecting `AKV_URLS` without `AKV_KEY_NAMES` in envconfig -- and dropping `name` from the log line.

cmd/webhook/main.go
```diff
@@ -91,7 +106,26 @@ func main() {
...
+		akvclient, err := azsecrets.NewClient(baseCfg.AKVUrls[0], cred, nil)
+		if err != nil {
+			log.Panicf("could not create client: %v", err)
+		}
+		for _, name := range strings.Split(webhookConfig.WebhookSecret, ",") {
+			resp, err := akvclient.GetSecret(ctx, name, "", nil)
+			if err != nil {
+				log.Panicf("error fetching webhook secret %s: %v", name, err)
+			}
+			if resp.Value == nil {
```
jmeridth: :robot: (medium) An empty vault secret (`az keyvault secret set --value ""`) passes this nil check and gets appended; go-github's `ValidatePayloadFromBody` skips HMAC validation entirely when the secret is empty, allowing unsigned webhook forgery. Reject empty too (the GCP branch above shares this latent gap):
```suggestion
			if resp.Value == nil || *resp.Value == "" {
```

pkg/envconfig/envconfig.go
```diff
@@ -86,14 +89,31 @@ func BaseConfig() (*EnvConfig, error) {
...
+	if len(cfg.AKVKeys) > 0 && len(cfg.AKVKeys) != len(cfg.AppIDs) {
+		return nil, fmt.Errorf("AKV_KEY_NAMES length (%d) must match GITHUB_APP_IDS length (%d)", len(cfg.AKVKeys), len(cfg.AppIDs))
+	}
+
+	if len(cfg.AKVKeys) > 0 && len(cfg.AKVUrls) == 0 {
+		return nil, errors.New("AKV_KEY_NAMES is set but AKV_URLS is empty")
+	}
+
+	if len(cfg.AKVUrls) > 1 && len(cfg.AKVUrls) != len(cfg.AppIDs) {
+		return nil, fmt.Errorf("AKVUrls length (%d) must match GITHUB_APP_IDS length (%d) or be a single vault", len(cfg.AKVUrls), len(cfg.AppIDs))
```
jmeridth: :robot: (low) `AKV_KEY_VERSIONS` is never length-validated (unlike `AKV_KEY_NAMES`/`AKV_URLS`). If it's shorter than `GITHUB_APP_IDS`, `cmd/app` bounds-checks the index so there's no panic -- but the unlisted apps silently sign with the vault's *latest* version, defeating version pinning with no startup error. Suggest mirroring the other checks: error when `len(AKVKeyVersions) > 0 && len(AKVKeyVersions) != len(AppIDs)`.

jmeridth:
:robot: AI-assisted review -- 4 models (Opus, Sonnet, Haiku, Fable).

Thanks @TomPeas! The core signing is correct -- AKV's digest-based RS256 flow (SHA-256 -> `azkeys.Sign` -> `base64.RawURLEncoding`) is handled right (nice catch on the digest asymmetry vs GCP), and credential/concurrency look good. Left a few inline notes: three mediums (ctx-on-SIGINT, webhook secret-source branching + a secret-to-logs case, empty-secret fail-open) and one low (version validation).

One coordination flag: this overlaps open #1141 (AWS KMS). Both add a signing backend and edit the same four wiring files (`cmd/app/main.go`, `cmd/webhook/main.go`, `envconfig.go`, `ghtransport.go`), but diverge structurally -- this mirrors `pkg/gcpkms`, while #1141 refactors signing into a `pkg/kms` provider factory. They'll conflict on merge, so worth a maintainer call on whether Azure should land as a `pkg/kms/azure` provider under that factory rather than a standalone package.

jmeridth:
:robot: Heads-up: #1141 just merged, which refactored signing into a `pkg/kms` provider factory (replacing `pkg/gcpkms`) and reworked `ghtransport.New`, `envconfig`, and both `main.go`s. This PR now conflicts and will need a rebase. Since it currently mirrors the old `pkg/gcpkms`, the cleanest path is to add Azure as a `pkg/kms/azure` provider under the new `NewKMS` factory rather than a standalone package -- that also resolves the wiring collision we flagged. Happy to re-review once it's rebased.

TomPeas:
> One coordination flag: this overlaps open #1141 (AWS KMS). Both add a signing backend and edit the same four wiring files (`cmd/app/main.go`, `cmd/webhook/main.go`, `envconfig.go`, `ghtransport.go`), but diverge structurally -- this mirrors `pkg/gcpkms`, while #1141 refactors signing into a `pkg/kms` provider factory. They'll conflict on merge, so worth a maintainer call on whether Azure should land as a `pkg/kms/azure` provider under that factory rather than a standalone package.

Thanks for the review @jmeridth! I have no strong opinions about the two KMS structures (perhaps slightly favouring the KMS factory) so would be happy to refactor this into that structure if the decision to go that way is taken.

pkg/azurekeyvault/azurekeyvault.go
TomPeas: Implemented

cmd/webhook/main.go
TomPeas: AKV specific env values have now been removed and KMSKeys is used for everything

cmd/webhook/main.go
TomPeas: Check implemented

pkg/envconfig/envconfig.go
TomPeas: AKV specific env values have now been removed and KMSKeys is used for everything

TomPeas:
@jmeridth I have refactored this implementation to fit in with the new KMS and Secret factories and have addressed your comments where relevant after the refactor. If I could get another review that would be great, thanks!

jmeridth (approved):

PR #1524, opened 2026-08-01
fix(octosts): detect typed GitHub rate-limit errors on trust policy reads

### LINKED ISSUE #1526: Trust policy reads misclassify GitHub rate limits, disabling stale-cache fallback and cross-app retry
## Summary

`fetchTrustPolicyRaw` misclassifies GitHub rate limits, so both of the resilience mechanisms that exist for that case are dead in production: the stale trust-policy cache is never served, and `lookupTrustPolicyWithRetry`'s cross-app retry never fires. Callers see `NotFound` where they should see `ResourceExhausted`.

## Root cause

`pkg/octosts/octosts.go`, in `fetchTrustPolicyRaw`:

```go
var ghErr *github.ErrorResponse
if errors.As(err, &ghErr) && ghErr.Response != nil {
	switch ghErr.Response.StatusCode {
	case http.StatusForbidden, http.StatusTooManyRequests:
		// serve stale, else ResourceExhausted
	case http.StatusNotFound:
		trustPolicies.Add(tpKey, negativeCacheConst)
	}
}
```

Real GitHub sends `X-RateLimit-Remaining: 0` on a primary rate limit. Under that header go-github's `CheckResponse` returns a **`*github.RateLimitError`**, not a `*github.ErrorResponse` — and the two types do **not** unwrap to each other. `errors.As(err, &ghErr)` is therefore `false`, the whole block is skipped, and execution falls through to the generic not-found path.

Secondary ("abuse") limits have the same problem via `*github.AbuseRateLimitError`.

Verified against go-github v88.

## Impact

For any org hitting a GitHub rate limit on a trust policy read:

- **No stale-cache fallback.** `staleTrustPolicies` exists precisely so a rate limit doesn't fail an exchange, and it is never consulted.
- **No cross-app retry.** `isRateLimit` is false, so a multi-app deployment never tries another App's installation even when it has quota.
- **Wrong status code.** The caller gets `codes.NotFound` ("unable to find trust policy") instead of `codes.ResourceExhausted`, which is misleading to operators and to any client retry logic keyed on the code.

There is also a latent cache-poisoning risk in the same block: because the 404 branch is a `case` in the same `switch`, a response that *is* rate-limited but arrives shaped as a bare `*ErrorResponse` with status 404 would be negative-cached for five minutes.

## Reproduction

The bug is invisible to the current test suite, which is worth knowing before trying to reproduce it. `newFakeGitHubRateLimit` in `pkg/octosts/octosts_test.go` returns a 403/429 **without** `X-RateLimit-Remaining`, which produces a bare `*ErrorResponse` — the one shape the buggy code handles. So `TestExchangeRateLimit`, `TestPolicyReadRetriesOnRateLimit`, `TestPolicyReadAllRateLimitedReturnsError`, and `TestRateLimitServesStaleCache` all pass against a broken classifier.

To make it visible, add the header that real GitHub sends to that fake's contents handler:

```go
w.Header().Set("X-RateLimit-Remaining", "0")
w.Header().Set("X-RateLimit-Limit", "5000")
w.Header().Set("X-RateLimit-Reset", fmt.Sprintf("%d", time.Now().Add(time.Minute).Unix()))
```

Then:

```
go test ./pkg/octosts/ -run 'TestExchangeRateLimit|TestPolicyRead|TestRateLimitServesStaleCache' -v
```

All four fail: the two status-code assertions report `NotFound` where `ResourceExhausted` was wanted, `TestPolicyReadRetriesOnRateLimit` fails because no retry happens, and `TestRateLimitServesStaleCache` fails because stale is never served.

Minimal standalone check of the underlying type behaviour, independent of this codebase:

```go
resp := &http.Response{StatusCode: 403, Header: http.Header{"X-Ratelimit-Remaining": []string{"0"}}}
err := &github.RateLimitError{Response: resp}
var e *github.ErrorResponse
fmt.Println(errors.As(err, &e)) // false
```

## Note on prior art in this repo

`pkg/webhook` has its own `isGitHubRateLimited` that already checks the typed errors first. Its comment says it is "mirroring pkg/octosts' generic 403/429 handling" — so the two implementations diverged, and the weaker one is the one that got copied into the path where it matters.

## Suggested fix

A single shared helper that checks `*github.RateLimitError` and `*github.AbuseRateLimitError` before falling back to status codes, used by both packages so they cannot drift again; and reordering so the rate-limit check precedes the 404 branch, so a rate limit can never be negative-cached.

salekseev:
Fixes #1526.

Split out of #901 at @jmeridth's request — independent of that feature, and cherry-pickable on its own.

See #1526 for the diagnosis, impact, and reproduction steps. In short: a real primary rate limit yields `*github.RateLimitError`, which does not unwrap to `*github.ErrorResponse`, so `fetchTrustPolicyRaw`'s rate-limit branch never ran in production — the stale-cache fallback and the cross-app retry were both dead, and callers got `NotFound` instead of `ResourceExhausted`.

## What this changes

**`IsGitHubRateLimited` in `pkg/octosts`**, checking the typed errors before falling back to status codes:

```go
var rateLimitErr *github.RateLimitError
var abuseRateLimitErr *github.AbuseRateLimitError
if errors.As(err, &rateLimitErr) || errors.As(err, &abuseRateLimitErr) {
	return true
}
// ...then 403/429 on a bare *ErrorResponse
```

It is deliberately **lenient** — a bare 403 with no rate-limit marker also counts. That is correct for both current callers, where a false positive costs a retry or a webhook redelivery rather than granting access. The doc comment says so explicitly, because anything that gates *access* needs a stricter test and should not reuse this.

**`pkg/webhook`'s local `isGitHubRateLimited` is deleted** and both of its call sites now use the shared helper. That copy already checked the typed errors correctly; its comment said it was "mirroring pkg/octosts' generic 403/429 handling", so the two had diverged and the weaker one was in the path that mattered. One implementation means they cannot drift again.

**Ordering:** the rate-limit check now runs *before* the 404 branch, so a rate-limited response can never be mistaken for a missing policy and negative-cached — which would have poisoned the cache for five minutes.

The `ResourceExhausted` message loses its `(%d)` status code, since with a typed `*RateLimitError` there is no `ghErr` to read it from. The status is already on the log line above.

## Reviewing the test change first

The commit adds `X-RateLimit-Remaining: 0` (plus `-Limit` and `-Reset`) to `newFakeGitHubRateLimit` **before** touching the classifier, because the fake's omission of that header is why four existing tests passed against a broken classifier. If you want to see the bug rather than take my word for it: check out this commit and revert just the body of `IsGitHubRateLimited` to a status-code-only check — `TestExchangeRateLimit`, `TestPolicyReadRetriesOnRateLimit`, `TestPolicyReadAllRateLimitedReturnsError`, and `TestRateLimitServesStaleCache` all go red.

### DIFF
--- a/pkg/octosts/octosts.go
+++ b/pkg/octosts/octosts.go
@@ -351,6 +351,38 @@ func isRateLimit(err error) bool {
 	return ok && st.Code() == codes.ResourceExhausted
 }
 
+// IsGitHubRateLimited reports whether err looks like a GitHub rate-limit
+// response, primary or secondary.
+//
+// The typed errors must be checked first: go-github returns *RateLimitError for
+// a 403 carrying X-RateLimit-Remaining: 0 and *AbuseRateLimitError for a
+// secondary limit, and NEITHER unwraps to *ErrorResponse. A status-code-only
+// check therefore misses every genuine rate limit from real GitHub — which is
+// the bug this function exists to fix.
+//
+// This is deliberately LENIENT: a bare 403 with no rate-limit marker also
+// counts. That is correct for both current callers, where a false positive
+// costs a retry or a redelivery rather than granting access. Anything that
+// gates access needs a stricter test.
+func IsGitHubRateLimited(err error) bool {
+	if err == nil {
+		return false
+	}
+	var rateLimitErr *github.RateLimitError
+	var abuseRateLimitErr *github.AbuseRateLimitError
+	if errors.As(err, &rateLimitErr) || errors.As(err, &abuseRateLimitErr) {
+		return true
+	}
+	var errResp *github.ErrorResponse
+	if errors.As(err, &errResp) && errResp.Response != nil {
+		switch errResp.Response.StatusCode {
+		case http.StatusForbidden, http.StatusTooManyRequests:
+			return true
+		}
+	}
+	return false
+}
+
 type trustPolicy interface {
 	Compile() error
 }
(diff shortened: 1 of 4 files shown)


jmeridth:
:robot: (low) AI-assisted review -- 2 models (Opus 4.8, Fable 5), verified vs go-github v88 + tests.

Clean fix. Confirmed the premise: a real primary rate limit is `*github.RateLimitError`, which doesn't unwrap to `*ErrorResponse`, so the old `errors.As(*ErrorResponse)` check missed every genuine rate limit -- the stale-cache fallback and cross-app retry were both dead. The typed check restores them.

Verified no regression: `fetchTrustPolicyRaw`'s 404 negative-caching is preserved and mutually exclusive with the rate-limit path (go-github gates the typed errors on 403/429 before header checks, so a 404 with rate headers still yields a plain 404); the de-dup of the webhook's `isGitHubRateLimited` (from #1474) into the exported `octosts.IsGitHubRateLimited` is behavior-preserving with pure call-site repoints; no import cycle; and the test fake now sends `X-RateLimit-Remaining: 0` so it genuinely produces `*RateLimitError`.

One thing worth a conscious sign-off (pre-existing, not introduced here): a genuine permission `403` is still classified as rate-limited -> serves a stale trust policy for up to the 1h TTL. It's bounded -- `Exchange` still has to mint a live installation token afterward, which fails for a truly revoked install. LGTM.

jmeridth (approved):

PR #1525, opened 2026-08-01
feat(ghinstall): add Manager.GetAll for exhaustive installation enumeration

### LINKED ISSUE #1527: Manager.Get cannot be used to enumerate installations, so callers needing all of them can reach false conclusions
## Summary

`ghinstall.Manager` has no way to ask "which installations exist for this owner?". `Get` looks like it could be used for that with a retry loop, but it is a *routing* primitive — it answers "which installation should serve this request?" — and repeated calls give no coverage guarantee. A caller that loops on `Get` to survey installations can observe the same one repeatedly and draw a false conclusion about the whole set.

## Why `Get` cannot enumerate

Three mechanics in `pkg/ghinstall/ghinstall.go`, all working as designed for routing:

**1. With warm quota data, selection is deterministic.** `roundRobin.Get` first calls `pickByQuota`. When a quota store is configured and has data for every candidate, `pickFromPool` selects `argmax(remaining)` within the highest non-empty tier (comfortable / tight / lastResort). There is no rotation, so consecutive calls with unchanged quota return the **same** installation.

**2. Without quota data, rotation exists but guarantees nothing.** If the quota store is unconfigured, or any candidate has no entry, `pickByQuota` bails and `Get` falls back to `rr.counter.Add(1) % len(rr.managers)`. That rotates, but it is a single shared atomic counter — concurrent callers perturb each other's sequence, so N calls do not cover N distinct installations.

**3. The multi-app fallback only fires on `NotFound`.**

```go
if st, ok := status.FromError(err); ok && st.Code() == codes.NotFound {
    // try the remaining apps
}
```

Any other error returns immediately. That includes the **422** GitHub returns when minting a token scoped to a repository an installation cannot access — the exact error a caller would hit while probing whether some installation can read a particular repo.

## Impact

Any caller whose correctness depends on having surveyed *all* installations can be handed a false answer. The concrete case: probing whether some installation can read an organization's `.github` repository. With warm quota data, a loop over `Get` can be answered by the same blind installation every time, and conclude "no installation can read `.github`" while another installation can read it perfectly well.

If that conclusion gates a security control, the false answer is a **fail-open** — and it is cached, so it persists past the moment.

## A related limitation any enumeration inherits

Worth stating alongside, because a `GetAll` built on the existing caches cannot be more authoritative than they are:

- the positive installation LRU has **no TTL**, so an App uninstalled from an org keeps appearing indefinitely;
- the negative cache hides a newly installed App until its entry expires.

So the honest guarantee for any enumeration over this layer is "every installation this process has observed", not "every installation that currently exists". Callers that need more than that need an authoritative source (`ListInstallations`) and invalidation on installation webhooks.

## Reproduction

Structural rather than a failing test today, since there is no caller to break. A `Manager` fake whose `Get` always returns `installs[0]` — which is exactly what `pickByQuota` does with unchanged quota — makes it observable:

```go
func (e *enumMgr) Get(context.Context, string, string, string) (*ghinstallation.AppsTransport, int64, error) {
	return e.installs[0].Transport, e.installs[0].ID, nil
}
```

Given two installations where the first cannot read the target repository and the second can, any number of `Get` calls sees only the first, and a survey built on it concludes the repository is unreadable.

## Suggested fix

An explicit `GetAll(ctx, owner) ([]Installation, error)` that performs no routing or capacity selection, with a contract that says plainly what it does and does not promise — in particular that a **non-nil error means the enumeration is not exhaustive even when the returned slice is non-empty**, so callers that need completeness for correctness treat it as unknown rather than as a complete answer.

salekseev:
Fixes #1527.

Split out of #901 at @jmeridth's request.

**Note up front: this adds a `Manager` method with no in-tree caller yet** — its consumer is the org trusted-issuer lookup that stays in #901. I split it out because the interesting part is the *contract*, which is easier to argue with on its own than buried in a 4K-line diff. If you'd rather not merge an unused method, say so and I'll fold it back.

See #1527 for why `Get` cannot be used to enumerate — briefly, `pickByQuota` selects `argmax(remaining)` with no rotation when quota data is warm, the counter fallback gives no coverage guarantee, and the multi-app fallback fires only on `codes.NotFound` so a 422 returns immediately.

## The contract, which is the substance of this PR

```go
GetAll(ctx context.Context, owner string) ([]Installation, error)
```

The doc comment is where the review value is. The parts worth pushing back on:

- **A non-nil error means the enumeration is NOT exhaustive, even when the returned slice is non-empty.** Partial results come back *alongside* the error, so a best-effort caller can use them — but a caller whose correctness depends on completeness must treat a non-nil error as unknown. Returning partial data and an error together is unusual enough that I wrote down why.
- **Exhaustiveness is bounded by `Get`'s cache**, and I documented that rather than let a caller assume otherwise: the positive LRU has no TTL, so an uninstalled App lingers indefinitely; the negative cache hides a newly installed App until its entry expires. The honest guarantee is "every installation this process has observed", not "every installation that exists". Anything needing more wants `ListInstallations` plus invalidation on installation webhooks — out of scope here, and called out in #1527.
- **Ordering is not stable across calls.** Whichever managers respond successfully occupy the early indices, so index 0 is not "the preferred App".
- **Callers must test `len()`, not nil-ness** — implementations differ on whether empty is nil or zero-length.

## Implementations

`manager.GetAll` (single app) translates `codes.NotFound` to `(nil, nil)`, since "not installed" is a complete answer rather than a failure.

`roundRobin.GetAll` concatenates in managers order, dedups on installation ID, checks `ctx.Err()` per iteration, and returns `codes.Unavailable` when any sub-manager failed — while still returning what it did collect. An earlier draft did `errs = append(...); continue` and dropped the partial results of a failing sub-manager, which contradicted the contract documented in the same function; there's a test for that now.

## Tests

`pkg/ghinstall/ghinstall_test.go` covers the dedup, partial-results-with-error, context cancellation, the `NotFound` → `(nil, nil)` translation, and the real unserved-owner path. `Installation` is also renamed from `Install` here, for readability at call sites.

Three fakes in `pkg/octosts/octosts_test.go` and one in `pkg/ghinstall/picker_test.go` gain the method. Worth knowing if you add fakes later: `fakePickerManager` claims interface conformance only in a comment, with no `var _ Manager = ...` assertion, so adding a method to `Manager` does not produce a compile error there. I found it by grepping, not by building.

### DIFF
--- a/pkg/ghinstall/ghinstall.go
+++ b/pkg/ghinstall/ghinstall.go
@@ -66,6 +66,9 @@ type Manager interface {
 	// currently exists. An App uninstalled from an org keeps appearing here
 	// indefinitely; a newly-installed App stays invisible until any negative
 	// cache entry for it expires.
+	//
+	// Callers must test len(), not nil-ness: implementations differ in whether
+	// an empty result is a nil or a zero-length slice.
 	GetAll(ctx context.Context, owner string) ([]Installation, error)
 }
 
@@ -332,7 +335,8 @@ func (rr *roundRobin) GetAll(ctx context.Context, owner string) ([]Installation,
 
 	if len(errs) > 0 {
 		err := status.Errorf(codes.Unavailable, "enumerating installations for %q: %v", owner, errors.Join(errs...))
-		clog.WarnContextf(ctx, "GetAll: enumeration incomplete, issuer allowlist evaluation may be unreliable: %v", err)
+		clog.WarnContextf(ctx, "ghinstall: GetAll enumeration incomplete for %q: %d of %d managers failed; "+
+			"callers requiring exhaustiveness must treat this as unknown: %v", owner, len(errs), len(rr.managers), err)
 		return out, err
 	}
 	return out, nil
--- a/pkg/ghinstall/ghinstall.go
+++ b/pkg/ghinstall/ghinstall.go
@@ -19,11 +19,11 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-// Install is one GitHub App installation for an owner.
-type Install struct {
+// Installation is one GitHub App installation for an owner.
+type Installation struct {
 	Transport *ghinstallation.AppsTransport
 	ID        int64 // installation ID
-	AppID     int64 // owning App ID, for logging and dedup
+	AppID     int64 // owning App ID, for logging
 }
 
 // Manager looks up GitHub App installations by owner.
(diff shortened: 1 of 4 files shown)


jmeridth:
:robot: (low) AI-assisted review -- 2 models (Opus 4.8, Fable 5), verified by code reading + go-github v88 + `go test -race`.

Clean, well-designed primitive (the contract doc is excellent). Verified: the not-installed-vs-enumeration-failed distinction is airtight (`NotFound` only from a negative-cache hit or a successful no-match walk; every API failure -> `codes.Internal`, so nothing transient masquerades as "not installed" -- `TestManagerGetAllPropagatesFailure` pins it); `roundRobin.GetAll` never collapses a failure to a nil error (partial results + `codes.Unavailable`); dedup is sound; concurrency clean (`-race` passes).

One load-bearing caveat for the #901 review -- not a defect here, it's documented in the contract: the "observed-only" bound means `GetAll` can return `(empty, nil)` while an installation actually exists for up to the 5-min negative-cache TTL (org installs an App just after a not-installed check) -- the same "incomplete-treated-as-complete" shape as the original fail-open. #901's consumer must treat `nil error + empty` as "none *observed*", treat per-installation read failures as unknown/deny, and accept or shrink that window for the security check.

Merging the unused method is fine -- self-contained, both implementers, full unit coverage, keeps #901 reviewable. LGTM.

jmeridth (approved):

PR #1529, opened 2026-08-01
refactor(octosts): route GitHub client construction through one constructor

### LINKED ISSUE #1528: Duplicated GitHub client construction makes a missing GHES base URL silently wrong
## Summary

Four places construct a `github.Client` with the same `WithTransport` + optional `WithEnterpriseURLs` block. Because the GHES base URL is applied per-construction-site rather than in one place, a new site is silently wrong by default — it compiles, it works against github.com, and it fails only on GHES.

## The duplication

- `pkg/octosts/octosts.go` (`fetchTrustPolicyRaw`)
- `pkg/ghinstall/ghinstall.go` (`m.baseURL`)
- `pkg/webhook/webhook.go` (`e.Transport.BaseURL`)

each carrying its own copy of:

```go
opts := []github.ClientOptionsFunc{github.WithTransport(atr)}
if baseURL != "" {
	opts = append(opts, github.WithEnterpriseURLs(baseURL, baseURL))
}
client, err := github.NewClient(opts...)
```

## Why this is more than style

A fourth site was added on a feature branch (#901) without the `WithEnterpriseURLs` option. The consequence was not a cosmetic bug: on a GHES deployment that read went to `api.github.com`, failed, was classified as "this installation cannot read the repository", and — because the caller treats that as true of every installation — permitted every issuer for the entire deployment. A fail-open on a security control, caused entirely by a copied block missing one option.

**It could not be caught by a test.** The test fakes inject a transport that rewrites requests to a local `httptest` server, so the client's base URL is never consulted. A client constructed without the enterprise option still reaches the fake, and the whole suite stays green. Verified: reverting the fix leaves every test passing.

That is the part worth fixing structurally — a defect class that is invisible to the test suite and only reachable in a deployment mode most contributors don't run.

## Suggested fix

One constructor per package that owns the base-URL decision, plus a test asserting the resulting `BaseURL()` for both the github.com and GHES cases, and a check that the constructor is the only one — since the wiring cannot otherwise be pinned by behaviour.

salekseev:
Fixes #1528.

Split out of #901 at @jmeridth's request. Pure refactor of pre-existing duplication in `pkg/octosts` — no behaviour change on github.com, and it closes a fail-open class on GHES.

See #1528 for why this matters more than style. Short version: a fourth copy of this block was added on #901 without `WithEnterpriseURLs`, which on a GHES deployment sent an allowlist read to `api.github.com`, and because an unreadable repository is classified as "this installation cannot see it", permitted every issuer for the whole deployment.

## What changes

`pkg/octosts/client.go`:

```go
func newGitHubClient(transport http.RoundTripper, baseURL string) (*github.Client, error)
```

`http.RoundTripper` rather than `*ghinstallation.AppsTransport` — it is what `github.WithTransport` takes, both call sites already satisfy it, and it lets the test pass `http.DefaultTransport` without standing up a signing key. It returns a plain error; the caller keeps its existing `status.Errorf(codes.Internal, ...)` wrapping verbatim.

`fetchTrustPolicyRaw` now calls it. That is the only construction site in this package, and the only behavioural change is that there is one fewer place to forget the option.

Deliberately **not** touching `pkg/ghinstall` or `pkg/webhook`. They are separate packages with their own base-URL sources (`m.baseURL`, `e.Transport.BaseURL`), and widening this across packages would make the diff harder to reason about than the problem warrants. Happy to follow up if you'd like them unified too.

## Two tests, because behaviour alone cannot pin this

**`TestNewGitHubClientBaseURL`** asserts the resulting base URL. The expected strings were determined empirically against go-github v88 rather than assumed, which taught me two things worth writing down: `BaseURL` is a *method* in v88, not a field, and `WithEnterpriseURLs` appends `/api/v3/` when the given URL lacks it.

| `baseURL` | `client.BaseURL()` |
|---|---|
| `""` | `https://api.github.com/` |
| `https://ghes.example.com` | `https://ghes.example.com/api/v3/` |
| `https://ghes.example.com/api/v3/` | `https://ghes.example.com/api/v3/` |
| `https://ghes.example.com/api/v3` | `https://ghes.example.com/api/v3/` |

The last two pin idempotence and trailing-slash normalisation, so nobody later "fixes" a doubled `/api/v3/api/v3/` that never existed.

**`TestGitHubClientHasOneConstructor`** is a convention test, and I want to flag it explicitly since it is unusual. It exists because the *wiring* cannot be pinned by behaviour: the fakes route requests through an injected transport, so a call site that bypasses the helper still passes every test. I confirmed that — reverting the call site to a direct `github.NewClient` leaves the whole suite green. So this test scans the package's non-test files for `github.NewClient(` and fails if one appears outside `client.go`. It also asserts that `client.go` still *contains* the call, so it cannot quietly become vacuous if the file is later refactored; I mutation-tested that branch specifically. If you'd rather enforce this with a lint rule than a test, I'm happy to switch.

## One rename you'll see in the diff

`octosts_test.go`'s helper `newGitHubClient(t, handler) *ghinstallation.AppsTransport` is renamed to `newAppsTransport` — it returns a transport, not a client, and it was occupying the name the production constructor wants. Mechanical, test-only.

### DIFF
--- a/pkg/octosts/client.go
+++ b/pkg/octosts/client.go
@@ -1,0 +1,31 @@
+// Copyright 2026 Chainguard, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package octosts
+
+import (
+	"net/http"
+
+	"github.com/google/go-github/v88/github"
+)
+
+// newGitHubClient builds a GitHub client that talks through transport to
+// baseURL, which is empty for github.com and set to an API endpoint for GitHub
+// Enterprise Server.
+//
+// Every client-construction site in this package must go through this. An
+// earlier revision of the org-allowlist read constructed its own client without
+// the enterprise option, which sent the read to api.github.com on a GHES
+// deployment; because an unreadable .github repository is treated as "this
+// installation cannot see the file", that silently disabled issuer enforcement
+// for the entire deployment.
+//
+// The returned error is plain: each caller wraps it in the form its own
+// signature requires.
+func newGitHubClient(transport http.RoundTripper, baseURL string) (*github.Client, error) {
+	opts := []github.ClientOptionsFunc{github.WithTransport(transport)}
+	if baseURL != "" {
+		opts = append(opts, github.WithEnterpriseURLs(baseURL, baseURL))
+	}
+	return github.NewClient(opts...)
+}
--- a/pkg/octosts/client_test.go
+++ b/pkg/octosts/client_test.go
@@ -1,0 +1,104 @@
+// Copyright 2026 Chainguard, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package octosts
+
+import (
+	"bytes"
+	"net/http"
+	"os"
+	"strings"
+	"testing"
+)
+
+// TestNewGitHubClientBaseURL pins the base URL every client in this package is
+// built with. The expected strings were determined empirically from go-github
+// v88 rather than assumed: WithEnterpriseURLs appends /api/v3/ when the given
+// URL does not already end in it, and always leaves a trailing slash.
+//
+// This exists because an earlier revision of the org-allowlist read built its
+// own client without the enterprise option, so on a GitHub Enterprise Server
+// deployment the read went to api.github.com, failed, and was classified as
+// "this installation cannot see .github" — which silently disabled issuer
+// enforcement for the whole deployment.
+func TestNewGitHubClientBaseURL(t *testing.T) {
+	for _, tc := range []struct {
+		name    string
+		baseURL string
+		want    string
+	}{{
+		name:    "empty base URL uses the github.com default",
+		baseURL: "",
+		want:    "https://api.github.com/",
+	}, {
+		name:    "enterprise host gains the /api/v3/ suffix",
+		baseURL: "https://ghes.example.com",
+		want:    "https://ghes.example.com/api/v3/",
+	}, {
+		name:    "enterprise API endpoint is preserved",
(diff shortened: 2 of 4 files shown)


jmeridth:
:robot: (low) AI-assisted review -- 2 models (Opus 4.8, Fable 5), verified vs go-github v88 source + `go test`.

Clean, behavior-preserving refactor. `newGitHubClient(atr, s.baseURL)` is byte-for-byte identical to the removed inline block for both github.com (transport only) and GHES (+`WithEnterpriseURLs`), same `codes.Internal` error wrap. The `newGitHubClient`->`newAppsTransport` rename is a required same-package fix with all call sites repointed, and `Revoke` correctly stays outside the invariant (raw `http.Request`, not a client). Nice tests -- the base-URL pin plus the self-documenting "one constructor" convention test with its guard-the-guard.

Two low, non-blocking notes: (1) the convention test is a tripwire, not an airtight invariant -- it matches the literal `github.NewClient(` and only scans `pkg/octosts`, so sibling packages (`webhook`, `ghinstall`, `prober`) that also build clients stay unguarded (currently correct); worth a follow-up if you want the guarantee to generalize. (2) Pre-existing (not this PR): the helper tolerates a bare-host `GITHUB_BASE_URL` (go-github appends `/api/v3/`) while `Revoke` uses it verbatim, so a bare-host config would read policies fine but send revocations to the wrong path -- fail-soft, and the README documents the `/api/v3` form.

LGTM.

jmeridth (approved):

PR #1531, opened 2026-08-01
fix(webhook): consolidate validated-path matching and filter the policy-directory scan

### LINKED ISSUE #1530: handleCheckSuite validates every file in .github/chainguard on the initial-commit path, including non-policies
## Summary

`handleCheckSuite`'s new-branch / initial-commit path lists `.github/chainguard` and validates **every** entry in it, with no filtering. Any non-policy file in that directory is therefore fetched and parsed as a `TrustPolicy`, which fails the `Trust Policy Validation` check run. Every other code path in the webhook filters to `*.sts.yaml`; this one does not.

## Where

`pkg/webhook/webhook.go`, in the branch taken when the check suite's before-SHA is the zero hash (a new branch, or the first commit):

```go
_, dirContents, _, err := client.Repositories.GetContents(ctx, owner, repo, ".github/chainguard", &github.RepositoryContentGetOptions{Ref: sha})
if err != nil {
	return nil, err
}
for _, file := range dirContents {
	files = append(files, file.GetPath())
}
```

Contrast with the four other sites (`handlePullRequest`, `handlePush`'s compare branch, and two more in `handleCheckSuite`), each of which guards on `filepath.Match(".github/chainguard/*.sts.yaml", ...)`.

## Impact

Add a `README.md` to `.github/chainguard` — documenting your policies, say — and the check run fails on the initial-commit path with a parse error against a file that was never a trust policy. Same for a `.gitkeep`, an `OWNERS`, or anything else.

It is latent rather than routine: it needs both a non-policy file in that directory *and* the zero-hash path (new branch or first commit), so a repository whose policies were added incrementally never hits it. That is also why it has gone unnoticed.

## Reproduction

The existing coverage does drive this branch — there is a default-branch check-suite test that goes through it — but its directory listing contains only a `.sts.yaml`, so it cannot see the bug. Add a second entry to that listing:

```
.github/chainguard/test.sts.yaml
.github/chainguard/README.md
```

and the check run fetches both and concludes `failure`:

```
fetched content paths = [.github/chainguard/README.md .github/chainguard/test.sts.yaml]
expected success, got failure (summary: Failed to validate trust policy.
  1 error occurred:
    * .github/chainguard/README.md: GET .../contents/.github/chainguard/README.md?ref=deadbeef: 404 [])
```

## Suggested fix

Filter this branch like every other one. Given the same glob is inlined at five sites, the fix worth making is one shared predicate that all of them use, so a sixth site cannot be added without it.

Note that fixing this **changes existing behaviour**: files in `.github/chainguard` that are not trust policies stop being fetched and validated on that path.

salekseev:
Fixes #1530.

Split out of #901 at @jmeridth's request. Two commits, reviewable separately: a no-behaviour-change refactor, then the fix.

## `e7775b0` — one predicate instead of five copies

The `.github/chainguard/*.sts.yaml` glob was inlined at four call sites plus a `filterSTSFiles` helper (a method that never used its receiver). Now:

```go
func isValidatedPath(path string) bool
func filterValidatedFiles(files []string) []string
```

`isValidatedPath` is the only `filepath.Match` left in `pkg/webhook/webhook.go`. Each converted site keeps its other conditions exactly — a couple also check `file.GetStatus() != "removed"`, and those are untouched. `TestFilterSTSFiles` is renamed and `TestFilesFromPushEvent` updated for the new signature; that churn is mechanical.

No `repo` parameter, deliberately. #901 adds a second, differently-scoped file to that directory and `isValidatedPath` will then need `repo` to decide whether it applies — but adding the parameter here would leave it unused, and `unparam` would rightly flag it. That's the next PR's job.

## `8f94cbc` — the fix, which changes behaviour

See #1530 for the full write-up. `handleCheckSuite`'s new-branch / initial-commit branch lists the policy directory rather than diffing it, and appended every entry unfiltered — so a `README.md` in `.github/chainguard` was fetched and parsed as a `TrustPolicy`, failing the check run.

**Reviewers: this is a behaviour change to an existing path.** Files in `.github/chainguard` that are not `*.sts.yaml` are no longer fetched or validated on that path. I believe that is what everyone already assumed was happening — the other four sites have always filtered — but it is a change and shouldn't slip past.

## On the tests

The bug lives at a *call site*, so a unit test on `filterValidatedFiles` would pass with the fix unimplemented. `TestCheckSuiteDefaultBranchSkipsNonPolicyFiles` drives the real handler through the zero-hash branch with a listing of `test.sts.yaml` + `README.md`, and asserts both which content paths were fetched and that the check run concludes `success`. Before the fix it fails with:

```
fetched content paths = [.github/chainguard/README.md .github/chainguard/test.sts.yaml], want [.github/chainguard/test.sts.yaml]
expected success, got failure (... * .github/chainguard/README.md: ... 404 [])
```

An existing test already drove that branch, but its listing contained only a `.sts.yaml`, so it was blind to the bug.

One test row worth explaining, since it looks redundant: `TestIsValidatedPath` asserts `.github/chainguard/config.yaml → false`. That row is what pins the `.sts.` infix. `README.md` matches neither `*.sts.yaml` nor a loosened `*.yaml`, so it cannot distinguish the two — loosening the glob is only caught by a `.yaml` file in the right directory. I had originally expected the `README.md` row to cover it; it doesn't.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -466,8 +466,15 @@ func (e *Validator) handleCheckSuite(ctx context.Context, cs checkSuite) (*githu
 		if err != nil {
 			return nil, err
 		}
+		// This branch lists the policy directory rather than diffing it, so the
+		// entries are everything the directory holds — not just trust policies.
+		// Filter here as every diff-based path already does, otherwise unrelated
+		// files (a README, a .gitkeep) get fetched and parsed as trust policies
+		// and fail the check run.
 		for _, file := range dirContents {
-			files = append(files, file.GetPath())
+			if isValidatedPath(file.GetPath()) {
+				files = append(files, file.GetPath())
+			}
 		}
 	} else {
 		resp, _, err := client.Repositories.CompareCommits(ctx, owner, repo, cs.GetCheckSuite().GetBeforeSHA(), sha, &github.ListOptions{})
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -340,7 +340,7 @@ func (e *Validator) handlePush(ctx context.Context, event *github.PushEvent) (*g
 			return nil, err
 		}
 		for _, file := range resp.Files {
-			if ok, err := filepath.Match(".github/chainguard/*.sts.yaml", file.GetFilename()); err == nil && ok {
+			if isValidatedPath(file.GetFilename()) {
 				if file.GetStatus() != "removed" {
 					files = append(files, file.GetFilename())
 				}
@@ -396,7 +396,7 @@ func (e *Validator) handlePullRequest(ctx context.Context, pr *github.PullReques
 		return nil, err
 	}
 	for _, file := range resp {
-		if ok, err := filepath.Match(".github/chainguard/*.sts.yaml", file.GetFilename()); err == nil && ok {
+		if isValidatedPath(file.GetFilename()) {
 			if file.GetStatus() != "removed" {
 				files = append(files, file.GetFilename())
 			}
@@ -475,7 +475,7 @@ func (e *Validator) handleCheckSuite(ctx context.Context, cs checkSuite) (*githu
 			return nil, err
 		}
 		for _, file := range resp.Files {
-			if ok, err := filepath.Match(".github/chainguard/*.sts.yaml", file.GetFilename()); err == nil && ok {
+			if isValidatedPath(file.GetFilename()) {
 				if file.GetStatus() != "removed" {
 					files = append(files, file.GetFilename())
 				}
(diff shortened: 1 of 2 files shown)


jmeridth:
:robot: (low) AI-assisted review -- 2 models (Opus 4.8, Fable 5), verified by code reading + `go test -race`; one confirmed the fix test fails when the fix is reverted.

Clean, approve-ready. The refactor is byte-equivalent (`isValidatedPath` == the old glob; all 4 sites keep their `GetStatus() != "removed"` guard; `filterValidatedFiles` == the receiver-less `filterSTSFiles`). The fix strictly narrows the new-branch dir-listing to actual `.sts.yaml` policies with no false negative -- the exchange fetches exactly `.github/chainguard/<identity>.sts.yaml` (always matches) and re-parses independently, so nothing bad becomes exchangeable. Bonus: it also removes a latent nil-pointer panic (pre-PR a subdirectory entry in that dir hit go-github's nil `GetContent()` deref). Well tested -- `TestCheckSuiteDefaultBranchSkipsNonPolicyFiles` fails when the fix is reverted.

One pre-existing follow-up (both flagged, not introduced here, don't block on it): `request.GetIdentity()` is only checked non-empty, so a `/` or `..` makes the exchange fetch a *nested* policy the webhook's top-level glob never validates -- a missing advisory check, not a bypass (exchange still parses + `CheckToken`). Sanitizing identity to a single path segment would be a nice hardening. Filed as a separate issue. LGTM.

jmeridth (approved):

PR #1543, opened 2026-08-05
feat(ghinstall): add GetAllFresh and let the cache pair own its invariant

salekseev:
Tracked by #1546. Split out of #901 so the feature there is easier to review on its own. This is the `pkg/ghinstall` half: an enumeration primitive plus a cache fix, both of which stand on their own merits and neither of which mentions trusted issuers.

**Merge #901 first, then this.** I originally claimed "either order" here and that was wrong — I verified it by merging them both ways locally. This PR adds `GetAllFresh` to the `ghinstall.Manager` interface; #901 adds two fakes implementing `Manager` (`enumMgr`, `notInstalledMgr`) in a file that does not exist on `main`, so this PR cannot stub them. Whichever merges second leaves `main` with `*enumMgr does not implement ghinstall.Manager (missing method GetAllFresh)`. It fails quietly: `go build ./...` passes either way, there is no textual conflict, and GitHub reports both MERGEABLE — only the test build breaks, so the first merge is green and the damage shows on the second.

**So this PR needs one small change before it merges:** once #901 is in, add those two stub methods to `pkg/octosts/org_trusted_issuers_store_test.go`. They cannot be added now — without this PR's interface method they are unused methods on unexported types, which fails `golangci-lint`. Reviewing the two independently is still fine; only the merge order is constrained.

The change that needs both PRs — a confirmation step calling `GetAllFresh` before concluding no installation can read `.github` — is deliberately held back as a follow-up rather than coupling them. **That means `GetAllFresh` merges here with no production caller**, tests only; #1546 tracks that gap and the fail-open it exists to close.

## The problem

`Manager.GetAll` is exhaustive only over *observed* state. `manager.GetAll` maps a `NotFound` to `(nil, nil)`, and a `NotFound` served from the negative (not-installed) cache is indistinguishable from a real one — so an App installed within the last `negativeTTL` is **absent from the enumeration with no error to show for it**. An incomplete result that reads as a complete one.

That is harmless for a caller that needs *some* installation. It is unsafe for one whose conclusion flips when an App is missing.

## `GetAllFresh`

Enumerates without consulting the negative cache, so a recently-installed App is visible. Contract:

- The positive cache **is** still consulted, and a hit short-circuits the walk — so after an uninstall/reinstall it can report a stale installation ID *instead of* the current one. That fails closed rather than open: the stale ID's token mint fails, which is a failure rather than an agreement.
- It bypasses **this process's** negative cache, not GitHub's replication. An installation GitHub has not yet propagated is invisible here too, so this narrows a window rather than eliminating one.
- Like `GetAll`, a non-nil error means the enumeration is **not** exhaustive even when the returned slice is non-empty, and callers test `len()` rather than nil-ness.
- It costs one `ListInstallations` walk per App with no positive cache entry, so it is not for hot paths.

It is on the `Manager` interface rather than behind a type assertion for the same reason `GetAll` is (#1525): a `Manager` that silently failed the assertion would quietly lose whatever precondition the caller was checking.

## The cache fix

`Get`'s success path did not clear the negative cache entry, and `Get` reads the negative cache **before** the positive one.

`manager` has no mutex, so a `Get` whose walk finished before an install can arm a negative entry *after* another goroutine recorded the installation, leaving a negative entry shadowing a positive one that already holds the truth — and `Get` returning `NotFound` for an installation it knows exists.

**Correcting an overstatement I made above:** I called this "a routing bug reachable on `main` today", which oversells it. Sequentially it is unconstructible on `main` — `Get` returns early on a negative-cache hit, and the only positive-cache write happens after that check, so a positive entry cannot be created while a negative one is live. It needs a concurrent interleaving, and `Remove`-before-`Add` narrows that window rather than closing it (the losing goroutine's write still lands last), which is what `setInstalled`'s own comment says. So on `main` alone this fix has little standalone value; its real role is as a **prerequisite for `GetAllFresh`**, the first path that bypasses the negative cache and can therefore observe and repair the shadowed state. That is why it ships here rather than separately.

All four cache writes now route through `setInstalled` / `setNotInstalled`, with `Remove` before `Add` so a concurrent reader sees neither entry and re-walks — returning the correct answer, just slower — rather than reading the stale negative.

**This narrows the window rather than closing it, and the doc comment says so explicitly.** Check/walk/commit is not atomic, so a `Get` whose walk finished earlier can still re-shadow via `setNotInstalled`; closing it needs per-owner serialization or versioned observations, neither of which belongs in this PR. The residual fails safe: a spurious `NotFound` rejects exchanges rather than granting anything.

I would rather ship an honest narrowing with the remaining gap named than a comment claiming an invariant the code does not hold. An earlier revision of this change did claim it, and an adversarial review caught it.

## Also here

`roundRobin.GetAll` and `GetAllFresh` share one `enumerate` helper, so the cancellation bail-out, the dedup-on-installation-ID, and the partial-result-plus-error contract cannot drift between them. `manager.Get`'s `ListInstallations` walk moved into `lookupInstallation`, which caches nothing and reports `found` separately from `err` — a completed walk that found nothing is a complete answer; a failed walk is not.

Test fakes across `pkg/ghinstall` and `pkg/octosts` gain the method because they implement `Manager`. The five that delegate to their own `GetAll` say so: those tests exercise routing, not cache freshness.

## Testing

`TestManagerGetAllFreshBypassesNegativeCache` is the regression test: arm the negative cache the way ordinary routing traffic does, install the App, then assert `GetAll` is *still* blind while reporting no error — the test fails as vacuous if the negative cache is not actually hiding the install — and that `GetAllFresh` is not fooled. It also asserts the subsequent `Get` recovers, which is what pins the shadowing repair.

`TestManagerGetAllFreshRepairsCoexistingCacheEntries` seeds the positive+negative coexistence state directly (in-package, via a type assertion) rather than racing to it, asserts routing really is broken in that state so it cannot go vacuous, then asserts the repair. It covers the resulting *state*, not the schedule that produces it — the interleaving itself is not deterministically testable without locking or injected synchronization, and I did not want a flaky test standing in for a real one.

Alongside those: positive-cache short-circuit (asserted by request count), a genuine miss still arming the negative entry, a failed walk propagating as non-`NotFound`, and `roundRobin.GetAllFresh` covering dedup, partial-plus-error, unserved owner, and cancellation.

`golangci-lint run ./...` reports zero issues; `go test ./...` passes; `go test -race` is clean on `pkg/ghinstall` and `pkg/octosts`. `GetAllFresh` (both implementations), `enumerate`, `cacheKey`, and `setInstalled` are at 100% statement coverage. The negative-entry removal was mutation-tested: deleting it fails two tests.

## Not in scope

Making enumeration authoritative — invalidating on installation webhooks so the positive cache cannot report an uninstalled App — is the real fix for the other half of `GetAll`'s staleness and is deliberately not here.

### DIFF
--- a/pkg/ghinstall/ghinstall.go
+++ b/pkg/ghinstall/ghinstall.go
@@ -65,11 +65,29 @@ type Manager interface {
 	// installation this process has observed, not every installation that
 	// currently exists. An App uninstalled from an org keeps appearing here
 	// indefinitely; a newly-installed App stays invisible until any negative
-	// cache entry for it expires.
+	// cache entry for it expires. That missing App is reported as a nil error, so
+	// an incomplete enumeration is indistinguishable from a complete one; a caller
+	// whose conclusion becomes UNSAFE when an App is missing must confirm with
+	// GetAllFresh.
 	//
 	// Callers must test len(), not nil-ness: implementations differ in whether
 	// an empty result is a nil or a zero-length slice.
 	GetAll(ctx context.Context, owner string) ([]Installation, error)
+
+	// GetAllFresh enumerates like GetAll but does NOT consult the negative
+	// (not-installed) cache, so an App installed within the last negativeTTL is
+	// visible. The positive cache IS still consulted, so after an
+	// uninstall/reinstall this can report a stale ID; that fails closed, since
+	// the stale ID's token mint fails rather than agreeing. It bypasses this
+	// process's negative cache, not GitHub's replication, so it narrows the
+	// window rather than eliminating it.
+	//
+	// Use it to CONFIRM an absence conclusion GetAll can only support over
+	// observed state. It costs one ListInstallations walk per App with no
+	// positive cache entry, so it does not belong on a hot path. Like GetAll, a
+	// non-nil error means the enumeration is NOT exhaustive even if the returned
+	// slice is non-empty, and callers must test len() rather than nil-ness.
+	GetAllFresh(ctx context.Context, owner string) ([]Installation, error)
 }
 
 const defaultNegativeTTL = 5 * time.Minute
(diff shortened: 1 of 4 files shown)


salekseev:
Superseded by #901, which now contains this work — closing per @jmeridth's call to fold it in ([comment](https://github.com/octo-sts/app/pull/901#discussion_r3722078164)).

Both commits from this PR are in #901 unchanged: `GetAllFresh` plus routing the manager's cache writes through `setInstalled`/`setNotInstalled`. Folding also removes the interface/fakes build break that made the two PRs unmergeable in either order — #1543 added `GetAllFresh` to `ghinstall.Manager` while #901 added fakes implementing `Manager` in a file that did not exist on `main`, so whichever merged second left `main` with a red test build and no textual conflict to warn anyone.

More importantly it means the enumeration primitive no longer ships without its caller. `confirmNoAccess` is in #901 too, so the control arrives with its own enforcement rather than in a bypassable state — which was the whole point of #1546.

Nothing here is abandoned; it is all reviewable in #901, as the first of seven commits.

PR #1551, opened 2026-08-06
fix(webhook): enrich request context with app and installation IDs

TomPeas:
## What

Call `ghtransport.EnrichContext` in the webhook's three event handlers (`handlePush`, `handlePullRequest`, `handleCheckSuite`) so the request context carries the GitHub App ID and installation ID, plus a regression test covering the label plumbing end to end.

## Why

The `github_rate_limit_*` metrics emitted by the webhook had an empty `app_id` label (rendering as null in dashboards).

`httpmetrics.WrapTransport` reads both labels off the request context:

```go
appID, _ := r.Context().Value(githubAppIDKey).(string)
installationID, _ := r.Context().Value(githubInstallationIDKey).(string)
```

Callers that don't set these produce time series with `app_id=""` / `installation_id=""`. `pkg/octosts` already enriches at both of its call sites, but `pkg/webhook` never did — so every webhook-originated GitHub API call was emitting unattributable rate-limit series. With multiple apps sharing quota, that makes it impossible to tell which app is consuming what.

## Scope

`ServeHTTP` dispatches to exactly three handlers — `check_run` events reuse `handleCheckSuite` via `fauxCheckSuite` — so enriching those three covers every event path. All downstream calls (`CompareCommits`, `ListFiles`, `GetContents`, and `handleSHA` → `validatePolicies` / `CreateCheckRun`) inherit the context.

This also labels the installation token exchange, not just the API calls: `ghinstallation.Transport.RoundTrip` threads `req.Context()` through `Token(ctx)` → `refreshToken(ctx)`, which reattaches it to the `POST /app/installations/{id}/access_tokens` request.

`e.Transport.AppID()` is the correct source — `cmd/webhook/main.go` builds a single transport from `AppIDs[0]`, so it cannot drift from the app actually making the request. The client LRU caches the client, not the context, so per-event labels stay correct across cached installations.

## Testing

`TestWebhookEnrichesMetricsContext` delivers a push event through the real handler stack and asserts the resulting `github_rate_limit_remaining` series is labelled with the app and installation IDs.

Two details make the assertion possible:

- `httpmetrics` only instruments requests whose URL host is `api.github.com`, so the transport keeps that host and a `redirectTransport` beneath the metrics wrapper clones the request onto the `httptest` server. Cloning rather than 	rewriting in place matters — `httpmetrics` re-reads `r.URL.Host` after the round trip returns.
- The context keys are unexported, so the test asserts on the emitted metric
  via `prometheus.DefaultGatherer` rather than inspecting the context.

Verified the test reproduces the bug: with the `EnrichContext` calls removed it
fails, and the transport logs the original symptom.

```
--- FAIL: TestWebhookEnrichesMetricsContext
    no github_rate_limit_remaining series labelled app_id=424242
    installation_id=999999: the webhook did not enrich the request context

github_api_call ... app_id="" installation_id="" rate_limit_remaining=4999
```

`go build ./...` and `go test ./...` pass.

`go.mod` promotes `prometheus/client_golang` and `prometheus/client_model` from
indirect to direct, since the test now imports them.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/yaml"
 
+	"github.com/octo-sts/app/pkg/ghtransport"
 	"github.com/octo-sts/app/pkg/octosts"
 )
 
@@ -298,6 +299,7 @@ func (e *Validator) handlePush(ctx context.Context, event *github.PushEvent) (*g
 	repo := event.GetRepo().GetName()
 	sha := event.GetAfter()
 	installationID := event.GetInstallation().GetID()
+	ctx = ghtransport.EnrichContext(ctx, e.Transport.AppID(), installationID)
 
 	// Skip if the organization is not in the list of organizations to validate.
 	if e.shouldSkipOrganization(owner) {
@@ -351,6 +353,7 @@ func (e *Validator) handlePullRequest(ctx context.Context, pr *github.PullReques
 	repo := pr.GetRepo().GetName()
 	sha := pr.GetPullRequest().GetHead().GetSHA()
 	installationID := pr.GetInstallation().GetID()
+	ctx = ghtransport.EnrichContext(ctx, e.Transport.AppID(), installationID)
 
 	// Skip if the organization is not in the list of organizations to validate.
 	if e.shouldSkipOrganization(owner) {
@@ -415,6 +418,7 @@ func (e *Validator) handleCheckSuite(ctx context.Context, cs checkSuite) (*githu
 	repo := cs.GetRepo().GetName()
 	sha := cs.GetCheckSuite().GetHeadSHA()
 	installationID := cs.GetInstallation().GetID()
+	ctx = ghtransport.EnrichContext(ctx, e.Transport.AppID(), installationID)
 
 	// Skip if the organization is not in the list of organizations to validate.
 	if e.shouldSkipOrganization(owner) {
--- a/pkg/webhook/webhook_test.go
+++ b/pkg/webhook/webhook_test.go
@@ -16,9 +16,11 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/http/httputil"
+	"net/url"
 	"os"
 	"path/filepath"
 	"slices"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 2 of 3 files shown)


jmeridth:
@TomPeas have a conflict to fix then we can approve and merge. LGTM

TomPeas:
@jmeridth Conflict fixed

jmeridth:
:robot: (summary) AI-assisted review (Fable 5), verified locally against main. The fix is correct and the regression test is genuine.

Traced the pipeline end to end: `EnrichContext` in the three handlers sets the app/installation IDs, and they propagate to the downstream GitHub calls and the installation-token exchange. The new test is non-vacuous: stubbing out the three `EnrichContext` calls makes it fail with the original symptom (`app_id=""`), and the labeled series returns once restored. `go build`, `go mod verify`, `go vet`, and the full `pkg/webhook` suite pass; the go.mod promotion is `tidy`-clean.

Two low, non-blocking notes inline. LGTM.

pkg/webhook/webhook_test.go
```diff
@@ -2116,3 +2120,179 @@ func TestCheckSuiteAllowlistInvalidConcludesFailure(t *testing.T) {
...
+	if !ok {
+		t.Fatalf("no github_rate_limit_remaining series labelled app_id=%d installation_id=%d: "+
+			"the webhook did not enrich the request context", appID, installationID)
+	}
+	if got != remaining {
+		t.Errorf("github_rate_limit_remaining = %v, want %v", got, float64(remaining))
+	}
+
+	// The bug's signature: requests attributed to no app at all.
+	if _, ok := gaugeValueForLabels(t, "github_rate_limit_remaining", map[string]string{
```
jmeridth, line 2292: :robot: (low, non-blocking) `gaugeValueForLabels` reads the process-global `prometheus.DefaultGatherer`, so this negative assertion is coupled to global metric state. It's fine today (nothing else in the package wraps the metrics transport, no `t.Parallel`, and the unique IDs disambiguate the positive check at :2279), but if a future test in this package ever routes an unenriched request through `metrics.WrapTransport` for the same owner, this check could fail spuriously. Non-blocking, just flagging the latent coupling for future test additions.

pkg/webhook/webhook.go
```diff
@@ -321,6 +322,7 @@ func (e *Validator) handlePush(ctx context.Context, event *github.PushEvent) (*g
 	repo := event.GetRepo().GetName()
 	sha := event.GetAfter()
 	installationID := event.GetInstallation().GetID()
+	ctx = ghtransport.EnrichContext(ctx, e.Transport.AppID(), installationID)
```
jmeridth, line 325: :robot: (low, non-blocking) Minor: `EnrichContext` runs before the `shouldSkipOrganization` / early-return checks below, so the app/installation labels attach to the context even on no-op paths where no GitHub request is made. Zero-cost and harmless (nothing reads the labels unless a request goes out), and the same holds in `handlePullRequest` (:379) and `handleCheckSuite` (:444). Flagging only for completeness, no change needed.

jmeridth (approved):

PR #1590, opened 2026-08-18
fix: list check_suite PR files from the PR's base repository

toshi38:
A check_suite event firing in a fork can carry an associated pull request that lives in the parent (base) repository — the PR number is only meaningful there. handleCheckSuite listed the PR's files against the event repo, so for fork PRs the call returned a 404 that surfaced as an HTTP 500 from the webhook, on every push to the PR's head branch.

Resolve the PR's home from the base-repo API URL carried in the payload (the nested repo object has no owner login, so the URL is parsed) and list files there. Same-repo PRs resolve to the event repo as before, and the head SHA still lives in the event repo, so content validation is unchanged.

If the base repo is genuinely out of the installation's reach, a 404 now skips that PR with a warning instead of failing the whole event: the base repository's own check_suite covers validating the PR.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"mime"
 	"net/http"
+	"net/url"
 	"path/filepath"
 	"strings"
 	"sync"
@@ -80,6 +81,25 @@ var prActionsThatChangeFiles = sets.New("opened", "synchronize", "reopened")
 // installationClientCacheSize matches the app's other LRUs (e.g. ghinstall,
 // octosts); entries are tiny and least-recently-used installations evict first.
 const installationClientCacheSize = 200
+
+// repoFromAPIURL extracts the owner and repository name from a GitHub API
+// repository URL such as https://api.github.com/repos/{owner}/{repo} (or the
+// GitHub Enterprise Server .../api/v3/repos/{owner}/{repo} variant).
+// check_suite payloads identify a pull request's base repository by API URL
+// only — the nested repo object carries no owner login.
+func repoFromAPIURL(apiURL string) (owner, repo string, ok bool) {
+	u, err := url.Parse(apiURL)
+	if err != nil {
+		return "", "", false
+	}
+	parts := strings.Split(strings.Trim(u.Path, "/"), "/")
+	for i := 0; i+2 < len(parts); i++ {
+		if parts[i] == "repos" {
+			return parts[i+1], parts[i+2], true
+		}
+	}
+	return "", "", false
+}
 
 func isBotSender(sender *github.User) bool {
 	return sender != nil && sender.Login != nil && strings.HasSuffix(sender.GetLogin(), "[bot]")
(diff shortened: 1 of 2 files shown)


pkg/webhook/webhook.go
```diff
@@ -514,8 +534,26 @@ func (e *Validator) handleCheckSuite(ctx context.Context, cs checkSuite) (*githu
...
-		resp, _, err := client.PullRequests.ListFiles(ctx, owner, repo, pr.GetNumber(), &github.ListOptions{})
+		// A PR raised from a fork lives in the base repository and its number
+		// is only meaningful there. When this event fires in the fork, listing
+		// files against the event repo 404s, so resolve the PR's home from the
+		// base repo API URL carried in the payload.
+		prOwner, prRepo := owner, repo
+		if o, r, ok := repoFromAPIURL(pr.GetBase().GetRepo().GetURL()); ok {
+			prOwner, prRepo = o, r
+		}
+		resp, _, err := client.PullRequests.ListFiles(ctx, prOwner, prRepo, pr.GetNumber(), &github.ListOptions{})
```
jmeridth, line 545: :robot: (medium)

Nice switch to the base repo here. One gap right below: the file-classification call still uses the event repo name, so it does not follow the switch:

line 560 — `isValidatedPath(repo, file.GetFilename(), e.policyRepo())` should be `isValidatedPath(prRepo, ...)`

`isValidatedPath` gates the org-allowlist path (`trusted-token-issuers.yaml`) on `EqualFold(repo, orgPolicyRepo)`. With `repo` (the fork) instead of `prRepo` (the base), a fork PR whose repo name differs from the policy repo (`.github`) drops that file, so `len(files) == 0` and no check run is created for an org trust-anchor change. Regular `*.sts.yaml` policies are unaffected (the glob is repo-agnostic) — only the org-allowlist path is name-gated.

It only bites a renamed same-org fork of the policy repo, and it is a lost pre-merge lint signal rather than an enforcement bypass (exchange-time re-reads the policy). Still worth the one-word fix since it matches the PR own intent — and neither new test reaches this branch, so a case over the org-allowlist path would lock it in.

pkg/webhook/webhook.go
```diff
@@ -514,8 +534,26 @@ func (e *Validator) handleCheckSuite(ctx context.Context, cs checkSuite) (*githu
...
 		if err != nil {
+			// The installation may still be unable to see the PR's home (e.g.
+			// the base repo belongs to another owner where this app is not
+			// installed). There is nothing to validate from this side — the
+			// base repo's own check_suite covers the PR — so skip it rather
+			// than fail the whole event with a 5xx.
+			var ghErr *github.ErrorResponse
+			if errors.As(err, &ghErr) && ghErr.Response != nil && ghErr.Response.StatusCode == http.StatusNotFound {
```
jmeridth, line 553: :robot: (low, non-blocking)

The 404-only skip is the right call. Noting the sibling case for later: a reachable-but-forbidden base repo (403 / secondary rate limit) still falls through to `return nil, err` and 5xxs the whole event — the same redelivery-loop this PR fixes, via a different status. Consistent with how `CompareCommits`/`GetContents` are handled today, so fine to leave; if you want to fully close it you could skip + warn on 403 too.

jmeridth:
:robot: (summary)

Reviewed with four models (Opus 4.8, Sonnet 5, Haiku 4.5, Fable 5), verification-first. Solid fix — base-repo resolution, the URL parser, and the 404 skip all check out (no SSRF, content still read from the event repo at the head SHA, safe nil/empty fallback, and the skip is not an enforcement bypass).

One thing to address before merge (inline): the file-classification call in the PR loop still uses the event repo name instead of the base, silently dropping org-allowlist validation for a renamed fork of the policy repo. One-word fix. Plus one low, non-blocking note on non-404 error handling.

jmeridth:
Converting to draft. @toshi38 please make changes and mark `Ready for review` when ready. Thank you.

phelian:
I talked with @toshi38 and we agreed I would take over this work. #1702 replaces this PR, addresses the org-allowlist classification review feedback, and is stacked on #1701. Please continue review on #1702.

toshi38:
Closing in favour of https://github.com/octo-sts/app/pull/1702

PR #1591, opened 2026-08-18
fix: webhook 500 on pushes that create a new ref with 20+ commits

toshi38:
Pushes that create a new ref carry the zero SHA as `before`. At 20+ commits the handler falls back to the Compare API, which 404s on the zero SHA, and the webhook returns 500. Under 20 commits the payload path works fine, so this only hits big pushes: repo imports, first push of a long-lived branch, etc.

Can't just use the payload for these since it's truncated at 20 commits and the missing commits may contain policy files that were never validated. So scan the policy directory at the pushed SHA instead, same as handleCheckSuite already does for zero-SHA check suites. A 404 for the directory just means no policies (logged), any other error still fails the delivery. Small pushes keep using the payload path.

Tests: no Compare call for a 20-commit new-ref push (the check run comes from the directory listing), missing directory is a no-op, non-404 scan errors still 500, and <20-commit new-ref pushes stay on the payload path.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -351,15 +351,27 @@ func (e *Validator) handlePush(ctx context.Context, event *github.PushEvent) (*g
 
 	// GitHub push payloads include up to 20 commits. When not truncated,
 	// use the payload directly to avoid a Compare API call.
-	//
-	// A push that creates a new ref carries the zero SHA as "before", which
-	// the Compare API rejects with a 404, so use the payload path there
-	// regardless of commit count. Policy files beyond the truncation point
-	// were validated when their commits were first pushed, and a brand-new
-	// default branch is covered by handleCheckSuite's zero-SHA directory scan.
-	if len(event.Commits) < 20 || event.GetBefore() == zeroHash {
+	switch {
+	case len(event.Commits) < 20:
 		files = e.filesFromPushEvent(repo, event)
-	} else {
+	case event.GetBefore() == zeroHash:
+		// A push that creates a new ref carries the zero SHA as "before",
+		// which the Compare API rejects with a 404 (previously surfacing as a
+		// webhook 500). The payload is also truncated, so commits beyond the
+		// first 20 could carry policy files never seen before. Scan the
+		// policy directory at the pushed SHA instead — the same approach as
+		// handleCheckSuite's zero-SHA path. A missing directory just means
+		// there are no policies to validate.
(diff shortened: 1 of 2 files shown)


pkg/webhook/webhook.go
```diff
@@ -351,9 +351,27 @@ func (e *Validator) handlePush(ctx context.Context, event *github.PushEvent) (*g
...
 		files = e.filesFromPushEvent(repo, event)
-	} else {
+	case event.GetBefore() == zeroHash:
+		// A new-ref push: the Compare API rejects the zero "before" SHA with
+		// a 404, and the truncated payload may omit policy files, so scan the
+		// policy directory at the pushed SHA instead. A missing directory
+		// means there are no policies to validate.
+		_, dirContents, resp, err := client.Repositories.GetContents(ctx, owner, repo, ".github/chainguard", &github.RepositoryContentGetOptions{Ref: sha})
+		if err != nil {
+			if resp == nil || resp.StatusCode != http.StatusNotFound {
```
jmeridth, line 364: :robot: (low, non-blocking)

Clean guard — network errors (`resp == nil`) and non-404s both still 500, only an explicit 404 is swallowed. One for later, not this PR: a rate-limited 403 also falls through to `return nil, err` → 500 → GitHub redelivery, which re-hammers the rate-limited API. Every `GetContents`/`CompareCommits` here shares that gap and `handleSHA` already avoids it, so it is a pre-existing consistency thing.

jmeridth:
:robot: (summary)

Reviewed with two models (Opus 4.8, Fable 5), verification-first; both ran the new tests at head `1d0e46b` (all pass). Clean — no blocking findings.

The fix is fail-safe: the dir-404 swallow can only mean "no policy dir at the pushed SHA" (not attacker-inducible; network/non-404 errors still 500), the directory scan is a safe superset mirroring `handleCheckSuite`'s zero-SHA path, deletion pushes can not reach the new branch, and the switch ordering keeps `<20`-commit new-ref pushes on the payload path. Tests pin all four behaviors.

One follow-up candidate (out of scope here): `handleCheckSuite`'s own zero-SHA directory scan still treats a dir-404 as a hard error, so an initial commit to a repo without `.github/chainguard` can still 500 that path — the same class of bug this PR fixes for pushes. Filed separately.

jmeridth:
Converting to draft. @toshi38 please make changes and mark `Ready for review` when ready. Thank you.

phelian:
I talked with @toshi38 and we agreed I would take over this work. #1701 replaces this PR and includes the new-ref fix along with complete policy discovery for large diffs and PRs. Please continue review on #1701.

toshi38:
Closing in favour of https://github.com/octo-sts/app/pull/1701

PR #1598, opened 2026-08-21
fix(provider): refuse OIDC discovery connections to non-public addresses

lcasassa:
## What

Constrains OIDC discovery to destinations that are routable on the public internet, so a discovery request cannot be steered at an address inside the deployment's own network.

## Why

`provider.Get` receives the issuer that `apiauth.ExtractIssuer` read out of the bearer token, and it dials that issuer *before* the token's signature has been checked. The value is therefore chosen by an unauthenticated caller.

`oidcvalidate.IsValidIssuer` already constrains the shape of the issuer — scheme, hostname characters, path, length. What it cannot do is say where the name points. `https://internal-service.example` and a public name whose A record is a private address are both syntactically valid, so discovery will connect to them.

This complements the input sanitisation added in v0.5.3 (CVE-2025-52477) rather than replacing it: that work constrained the string, this constrains the destination.

## How

The HTTP client used for discovery now carries a dialer with a `Control` hook that rejects any address which is not publicly routable — loopback, RFC 1918, RFC 4193 unique local, link-local (where cloud instance metadata services live), carrier-grade NAT, the TEST-NETs, and the other reserved ranges. IPv4-mapped IPv6 addresses are unmapped first so `::ffff:10.0.0.1` is judged as the private address it is.

`Control` is the useful place for this. It runs after name resolution and immediately before `connect(2)`, on the concrete address being dialed, so a name that resolves inward is caught even if it resolved differently a moment earlier when it was validated as a string.

Two consequences worth noting:

- The constructed provider retains this client, so the JWKS fetches that follow are governed by the same rule. A discovery document pointing `jwks_uri` at an internal address is refused on the same terms as the issuer.
- Issuers that explicitly name a loopback host keep the relaxed treatment, preserving the local-development carve-out `IsValidIssuer` already makes for `localhost`, `127.0.0.1` and `::1`. A public hostname that merely *resolves* to loopback does not qualify.

## Testing

- `isPublicAddress` is table-tested across loopback, private, link-local, ULA, CGNAT, reserved, multicast, unspecified, IPv4-mapped and public addresses.
- The dialer is exercised against a real listener, asserting refusal by default and success when the issuer is itself loopback.
- The discovery transport is driven end to end to confirm a strict-mode client cannot reach a non-public address.

The existing `httptest`-based provider tests are unmodified and still pass: they use `127.0.0.1` issuers, which take the loopback carve-out.

`go build ./...`, `go vet ./pkg/provider/` and `go test ./pkg/...` are clean.

### DIFF
--- a/pkg/provider/dial.go
+++ b/pkg/provider/dial.go
@@ -1,0 +1,147 @@
+// Copyright 2026 Chainguard, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package provider
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"net/http"
+	"net/netip"
+	"net/url"
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/chainguard-dev/terraform-infra-common/pkg/httpmetrics"
+)
+
+// dialTimeout bounds how long a single discovery connection may spend in
+// connect(2). Discovery targets are attacker-influenced, so a target that
+// accepts packets but never completes the handshake must not pin a request.
+const dialTimeout = 5 * time.Second
+
+// errNonPublicAddress is returned when a discovery connection resolves to an
+// address that is not on the public internet.
+var errNonPublicAddress = errors.New("refusing to dial non-public address")
+
+// reservedPrefixes are ranges that are not routable on the public internet but
+// which netip's classifiers do not already cover. Cloud instance metadata
+// services live in the link-local range, which IsLinkLocalUnicast catches.
+var reservedPrefixes = []netip.Prefix{
+	netip.MustParsePrefix("0.0.0.0/8"),       // "this network" (RFC 1122)
+	netip.MustParsePrefix("100.64.0.0/10"),   // carrier-grade NAT (RFC 6598)
+	netip.MustParsePrefix("192.0.0.0/24"),    // IETF protocol assignments
+	netip.MustParsePrefix("192.0.2.0/24"),    // TEST-NET-1
+	netip.MustParsePrefix("198.18.0.0/15"),   // benchmarking (RFC 2544)
+	netip.MustParsePrefix("198.51.100.0/24"), // TEST-NET-2
+	netip.MustParsePrefix("203.0.113.0/24"),  // TEST-NET-3
+	netip.MustParsePrefix("240.0.0.0/4"),     // reserved (RFC 1112)
+	netip.MustParsePrefix("100::/64"),        // discard-only (RFC 6666)
+	netip.MustParsePrefix("2001:db8::/32"),   // documentation (RFC 3849)
+}
+
+// isPublicAddress reports whether addr is routable on the public internet.
+//
+// IPv4-mapped IPv6 addresses are unmapped first so that, for example,
+// ::ffff:10.0.0.1 is classified as the private address it really is.
+func isPublicAddress(addr netip.Addr) bool {
+	addr = addr.Unmap()
+	if !addr.IsValid() {
+		return false
+	}
+	// IsPrivate covers RFC 1918 (IPv4) and RFC 4193 unique local (IPv6).
+	if addr.IsLoopback() || addr.IsPrivate() || addr.IsUnspecified() ||
+		addr.IsLinkLocalUnicast() || addr.IsLinkLocalMulticast() ||
+		addr.IsInterfaceLocalMulticast() || addr.IsMulticast() {
+		return false
+	}
+	for _, p := range reservedPrefixes {
+		if p.Contains(addr) {
+			return false
+		}
+	}
+	return true
+}
+
+// newRestrictedDialer returns a dialer that refuses to open a connection to an
+// address that is not on the public internet.
+//
+// The check runs in Control, which the runtime calls after name resolution and
+// immediately before connect(2), on the concrete address being dialed. That
+// placement matters: validating the issuer string alone cannot see where a
+// hostname actually points, and a name that resolves differently between
+// validation and dial (DNS rebinding) would slip past a string check.
+//
+// allowLoopback relaxes the rule for loopback destinations only. It is set
+// solely for issuers that are themselves explicitly loopback, preserving the
+// local-development and test carve-out that oidcvalidate.IsValidIssuer makes.
+func newRestrictedDialer(allowLoopback bool) *net.Dialer {
+	return &net.Dialer{
+		Timeout:   dialTimeout,
+		KeepAlive: 30 * time.Second,
+		Control: func(_, address string, _ syscall.RawConn) error {
+			host, _, err := net.SplitHostPort(address)
+			if err != nil {
+				return fmt.Errorf("parsing dial address %q: %w", address, err)
+			}
+			addr, err := netip.ParseAddr(host)
+			if err != nil {
+				return fmt.Errorf("parsing dial address %q: %w", host, err)
+			}
+			if allowLoopback && addr.Unmap().IsLoopback() {
+				return nil
+			}
+			if !isPublicAddress(addr) {
+				return fmt.Errorf("%w: %s", errNonPublicAddress, addr)
+			}
+			return nil
+		},
+	}
+}
+
+// newDiscoveryTransport builds the RoundTripper used for OIDC discovery and,
+// because the constructed provider retains this client, for the JWKS fetches
+// that follow. A discovery document that points its jwks_uri at an internal
+// address is therefore refused on the same terms as the issuer itself.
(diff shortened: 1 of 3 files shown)


pkg/provider/dial.go
```diff
@@ -0,0 +1,147 @@
...
+const dialTimeout = 5 * time.Second
+
+// errNonPublicAddress is returned when a discovery connection resolves to an
+// address that is not on the public internet.
+var errNonPublicAddress = errors.New("refusing to dial non-public address")
+
+// reservedPrefixes are ranges that are not routable on the public internet but
+// which netip's classifiers do not already cover. Cloud instance metadata
+// services live in the link-local range, which IsLinkLocalUnicast catches.
+var reservedPrefixes = []netip.Prefix{
```
jmeridth, line 32: :robot: (medium)

This is a genuinely thorough denylist — metadata, RFC1918, ULA, CGNAT, TEST-NETs, broadcast all covered, and I verified each with a netip probe. One class slips through: IPv6 forms that embed an IPv4 address. `Unmap()` only strips the v4-mapped `::ffff:` form, so these classify as public:

- `64:ff9b::a00:1` (NAT64 well-known prefix, RFC 6052, embeds 10.0.0.1) — the credible one: on an IPv6-only / DNS64-NAT64 egress path, `iss = https://[64:ff9b::a9fe:a9fe]/` translates to 169.254.169.254 and reaches metadata past the guard.
- `::a00:1` (v4-compatible), `2002:a00:1::` (6to4) — same class, but deprecated / not routed on modern Linux, so theoretical.

Not reachable on the current Cloud Run path (no NAT64), so this is defense-in-depth rather than a live hole, but cheap to close by adding the prefixes:

```suggestion
var reservedPrefixes = []netip.Prefix{
	netip.MustParsePrefix("0.0.0.0/8"),       // "this network" (RFC 1122)
	netip.MustParsePrefix("100.64.0.0/10"),   // carrier-grade NAT (RFC 6598)
	netip.MustParsePrefix("192.0.0.0/24"),    // IETF protocol assignments
	netip.MustParsePrefix("192.0.2.0/24"),    // TEST-NET-1
	netip.MustParsePrefix("198.18.0.0/15"),   // benchmarking (RFC 2544)
	netip.MustParsePrefix("198.51.100.0/24"), // TEST-NET-2
	netip.MustParsePrefix("203.0.113.0/24"),  // TEST-NET-3
	netip.MustParsePrefix("240.0.0.0/4"),     // reserved (RFC 1112)
	netip.MustParsePrefix("100::/64"),        // discard-only (RFC 6666)
	netip.MustParsePrefix("2001:db8::/32"),   // documentation (RFC 3849)
	netip.MustParsePrefix("64:ff9b::/96"),    // NAT64 well-known (RFC 6052)
	netip.MustParsePrefix("64:ff9b:1::/48"),  // NAT64 local-use (RFC 8215)
	netip.MustParsePrefix("2002::/16"),       // 6to4
	netip.MustParsePrefix("::/96"),           // IPv4-compatible (deprecated)
}
```

Alternatively, inverting to an allowlist (reject unless `addr.IsGlobalUnicast() && !addr.IsPrivate()` and not reserved) closes this whole class including any exotic range added later.

pkg/provider/dial.go
```diff
@@ -0,0 +1,147 @@
...
+)
+
+// dialTimeout bounds how long a single discovery connection may spend in
+// connect(2). Discovery targets are attacker-influenced, so a target that
+// accepts packets but never completes the handshake must not pin a request.
+const dialTimeout = 5 * time.Second
+
+// errNonPublicAddress is returned when a discovery connection resolves to an
+// address that is not on the public internet.
+var errNonPublicAddress = errors.New("refusing to dial non-public address")
```
jmeridth, line 27: :robot: (medium)

`errNonPublicAddress` is a deterministic, permanent verdict, but it flows back through `newProviderWithRetry`, where the backoff loop only treats HTTP-status errors (via `isPermanentError`) as permanent — so a non-public address gets retried. On main the retry budget defaults to ~15 minutes (backoff v5 default `MaxElapsedTime`), so one request naming a host that resolves to 10.0.0.1 burns ~10 pointless attempts; an attacker just points an A record inward to trigger it at will.

(#1599 caps this to ~3 attempts / ~18s, but retrying a deterministic address rejection is wasted work either way.) Worth short-circuiting it as permanent at the retry site (in `newProviderWithRetry`'s `operation`):

```go
if errors.Is(err, errNonPublicAddress) {
    return nil, backoff.Permanent(err)
}
```

pkg/provider/dial.go
```diff
@@ -0,0 +1,147 @@
...
+		},
+	}
+}
+
+// newDiscoveryTransport builds the RoundTripper used for OIDC discovery and,
+// because the constructed provider retains this client, for the JWKS fetches
+// that follow. A discovery document that points its jwks_uri at an internal
+// address is therefore refused on the same terms as the issuer itself.
+func newDiscoveryTransport(allowLoopback bool) http.RoundTripper {
+	t := http.DefaultTransport.(*http.Transport).Clone()
```
jmeridth, line 109: :robot: (low, non-blocking)

`http.DefaultTransport.Clone()` carries `Proxy: ProxyFromEnvironment`. If an `HTTP(S)_PROXY` env var is ever set on the pod, the dialer connects to the proxy (a public address `Control` happily allows) and the real target is reached by the proxy via CONNECT — the IP guard is structurally bypassed. No proxy is set in the terraform today, so this is latent, but for a security control it is worth making un-disableable by a future env change:

```suggestion
	t := http.DefaultTransport.(*http.Transport).Clone()
	t.Proxy = nil
```

pkg/provider/dial.go
```diff
@@ -0,0 +1,147 @@
...
+
+// Transports are built once and shared so that connection pooling still
+// applies across discovery requests.
+var (
+	strictTransport   = sync.OnceValue(func() http.RoundTripper { return newDiscoveryTransport(false) })
+	loopbackTransport = sync.OnceValue(func() http.RoundTripper { return newDiscoveryTransport(true) })
+)
+
+// discoveryTransport returns the transport to use for the given issuer.
+func discoveryTransport(issuer string) http.RoundTripper {
```
jmeridth, line 122: :robot: (low, non-blocking)

The strict-vs-loopback transport choice is derived from the attacker-controlled issuer string, so a caller can opt into the relaxed policy by naming a loopback issuer (`http://127.0.0.1:<port>/...` passes `IsValidIssuer`). The carve-out correctly still blocks non-loopback internal addresses (verified), so the blast radius is confined to in-pod loopback — but that is non-empty (the `:2112` metrics listener), giving a pre-auth blind GET / port-scan-by-timing against loopback services. Since the carve-out exists for local dev/tests, gating it on deployment config (env or build flag) rather than the request would keep prod strict unconditionally while preserving the test path. Minor: `isLoopbackIssuer` also treats all of `127.0.0.0/8` as loopback, a touch wider than the exact-host set `IsValidIssuer` carves out.

jmeridth:
:robot: (summary)

Reviewed with four models (Opus 4.8, Sonnet 5, Haiku 4.5, Fable 5), verification-first, plus an independent netip classification probe and a trace through go-oidc v3.20.0. Really nice work on the approach — putting the check in `net.Dialer.Control`, on the resolved IP immediately before connect(2), is the right call and genuinely closes the validate-string-then-resolve (DNS-rebinding) gap that a string check cannot. The parts that matter most all hold: JWKS key fetches reuse the same guarded client, redirects are double-guarded (string + IP), every candidate IP is checked under Happy Eyeballs, parse failures fail closed, and the loopback carve-out does not extend to non-loopback addresses.

Two things worth fixing before merge (inline):
- The range list misses IPv6 forms that embed an IPv4 address (NAT64 `64:ff9b::/96` is the credible one — reaches internal v4/metadata on an IPv6-only/DNS64 egress path). Verified classifier gap; reachability is deployment-conditional, not current prod. Trivial to close, or invert to an allowlist.
- `errNonPublicAddress` is retried rather than treated as permanent, so a private-resolving issuer burns the full retry budget (~15 min on main) failing identically.

Two low, non-blocking notes inline (the cloned transport keeps `Proxy: ProxyFromEnvironment`; the loopback carve-out is selected from the attacker-supplied issuer string). And two test gaps worth adding: a case asserting a loopback issuer is still refused a non-loopback destination, and cases for the NAT64/transition ranges.

This composes with #1599 — that PR bounds the retry loop, which softens the second point above, though classifying the error permanent is still the right call. Thanks for tightening this surface; the core design is solid and these are completeness gaps, not a flaw in the approach.

jmeridth:

PR #1599, opened 2026-08-21
fix(provider): bound the cost of OIDC discovery

lcasassa:
## What

Gives OIDC discovery its own time and attempt budget instead of letting it run for as long as the caller is willing to wait.

## Why

`newProviderWithRetry` is reached with an issuer taken from a bearer token that has not been verified yet, so the work it performs is chosen by an unauthenticated caller. At present nothing bounds that work:

- each attempt inherits the caller's context, so a single attempt lasts as long as the client waits;
- `backoff.Retry` is given a `BackOff` but no `WithMaxTries` and no `WithMaxElapsedTime`, so it retries until the context is done.

A target that accepts the connection and then never responds therefore holds the request open for the caller's full deadline, issuing a fresh outbound request each round. The retry loop turns one inbound request into several outbound ones and multiplies the time each is held.

## How

- A per-attempt timeout, so one unresponsive target cannot consume the whole request.
- `WithMaxTries`, so the number of outbound requests per inbound request is fixed.
- `WithMaxElapsedTime`, so the attempts together are bounded.

The backoff schedule itself is unchanged, and a caller with a shorter deadline still wins — the per-attempt context derives from the caller's.

Cancelling the per-attempt context once an attempt returns is safe: go-oidc stores the HTTP client taken from the context, not the context itself (`oidc.go`, where the provider builds its `commonRemoteKeySet` from a fresh `context.Background()`), so neither the returned provider nor its later key set fetches depend on the attempt context.

The bounds are package-level variables rather than constants purely so tests can shorten them.

## Testing

- A new test asserts a failing target is retried a bounded number of times even when the caller's context is generous, so the bound demonstrably comes from the retry configuration and not from the caller giving up.
- A second asserts a target that accepts the connection and stalls cannot hold the call for the caller's deadline.

The existing retry tests are unmodified and still pass, including the ones asserting exactly three attempts and a minimum backoff delay.

`go build ./...`, `go vet ./pkg/provider/` and `go test ./pkg/...` are clean.

### DIFF
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -68,13 +68,36 @@ func Get(ctx context.Context, issuer string) (provider VerifierProvider, err err
 	return provider, nil
 }
 
+// Discovery is driven by an issuer taken from an unverified bearer token, so
+// its cost has to be bounded rather than left to the caller's deadline. A
+// target that stalls — accepting the connection and never responding — would
+// otherwise hold a request open for as long as the client waits, once per
+// attempt, with no ceiling on the number of attempts.
+//
+// These are variables rather than constants only so that tests can shorten
+// them; nothing outside this package changes them.
+var (
+	// discoveryAttemptTimeout bounds a single discovery attempt.
+	discoveryAttemptTimeout = 5 * time.Second
+	// discoveryMaxElapsedTime bounds all attempts together.
+	discoveryMaxElapsedTime = 15 * time.Second
+	// discoveryMaxTries bounds how many attempts are made.
+	discoveryMaxTries uint = 3
+)
+
 // newProviderWithRetry creates a new OIDC provider with exponential backoff retry logic
 func newProviderWithRetry(ctx context.Context, issuer string) (VerifierProvider, error) {
 	attempt := 0
 
 	operation := func() (VerifierProvider, error) {
 		attempt++
-		p, err := oidc.NewProvider(ctx, issuer)
+		// Bound the attempt itself. go-oidc keeps only the HTTP client from
+		// this context, not the context, so cancelling it here does not
+		// affect the returned provider or its later key set fetches.
+		attemptCtx, cancel := context.WithTimeout(ctx, discoveryAttemptTimeout)
+		defer cancel()
+
+		p, err := oidc.NewProvider(attemptCtx, issuer)
 		if err != nil {
 			clog.WarnContext(ctx, "provider creation failed", "attempt", attempt, "issuer", issuer, "error", err)
 			// Check for permanent errors that shouldn't be retried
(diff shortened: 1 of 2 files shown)


pkg/provider/provider.go
```diff
@@ -68,13 +68,36 @@ func Get(ctx context.Context, issuer string) (provider VerifierProvider, err err
...
+// target that stalls — accepting the connection and never responding — would
+// otherwise hold a request open for as long as the client waits, once per
+// attempt, with no ceiling on the number of attempts.
+//
+// These are variables rather than constants only so that tests can shorten
+// them; nothing outside this package changes them.
+var (
+	// discoveryAttemptTimeout bounds a single discovery attempt.
+	discoveryAttemptTimeout = 5 * time.Second
+	// discoveryMaxElapsedTime bounds all attempts together.
```
jmeridth, line 82: :robot: (low, non-blocking)

Careful work here — the per-attempt timeout plus `MaxTries` is exactly the right shape, and the note about go-oidc keeping only the HTTP client (not the context) checks out against v3.20.0, so the `defer cancel()` is safe.

One precision nit on this comment: with `discoveryMaxTries = 3` the elapsed guard never actually fires. backoff checks `MaxElapsedTime` only before a wait (`time.Since(start)+next > max`), so after attempt 2 (~11s) the check is ~13s < 15s and it proceeds to attempt 3 — `MaxTries` returns first. The real ceiling is `tries × (attemptTimeout + backoff)` ≈ 18s, not 15s. Still well-bounded; just worth making the comment match, or bumping `MaxTries` / lowering the per-attempt so 15s becomes the true governor.

```suggestion
	// discoveryMaxElapsedTime is a backstop on total elapsed time. With
	// discoveryMaxTries=3 the attempt cap governs first, so the effective
	// ceiling is tries*(attemptTimeout+backoff) ~= 18s.
```

jmeridth:
:robot: (summary)

Reviewed with two models (Opus 4.8, Fable 5), verification-first, tracing the go-oidc v3.20.0 + backoff v5 source directly. Thanks for this — well-reasoned and a real improvement: discovery goes from unbounded (one inbound request held for the caller's full deadline, once per unbounded retry) to a fixed 3 attempts with a per-attempt timeout.

The central safety claim holds: `oidc.NewProvider` stores only the HTTP client from the context, and `Provider.Verifier` builds its key set from a background context, so cancelling the per-attempt context can not affect the cached provider or later JWKS fetches. Confirmed in source by both models.

One low, non-blocking note inline on the ~18s vs 15s ceiling. Two optional follow-ups, out of scope here: (1) the "verification still works after the attempt context is cancelled" guarantee is argued in prose but not tested — a small test would lock it against a future go-oidc bump; (2) no negative caching / single-flight for a stalled issuer, so N concurrent requests for the same bad issuer each pay the full bounded cost (aggregate surface still linear in inbound requests). Filed the latter as a follow-up.

jmeridth:

PR #1601, opened 2026-08-21
feat(octosts): add OCTOSTS_ALLOWED_ISSUERS to confine OIDC discovery

lcasassa:
## What

Adds `OCTOSTS_ALLOWED_ISSUERS`, an optional list of issuers a deployment will perform OIDC discovery against. Unset — the default — permits any issuer, so nothing changes for an existing deployment until it is set.

## Why

A token names its own issuer, and that issuer's discovery document is what tells us the keys that verify the token. Discovery therefore has to happen *before* the signature can be checked. `Exchange` reflects that:

```go
issuer, err := apiauth.ExtractIssuer(bearer)   // from an unverified token
if !oidcvalidate.IsValidIssuer(issuer) { ... } // shape only
p, err := provider.Get(ctx, issuer)            // connects to issuer
tok, err := verifier.Verify(ctx, bearer)       // signature checked here
```

So a caller who presents nothing but an unsigned assertion chooses a host the service connects to, and learns from the timing and from its own logs whether the connection was made. The assertion never has to be valid.

The trust policy and the org-level allowlist both settle the issuer question properly, but they are consulted in `lookupInstallAndTrustPolicy`, after the provider has been fetched. The org allowlist additionally permits everything when an organization has not published one, which is the default. Neither constrains the request discovery has already made.

`oidcvalidate.IsValidIssuer` constrains the issuer's shape, which is worth doing, but any well-formed `https` URL satisfies it.

## How

The list is checked immediately after the format check and before `provider.Get`, so an issuer that is not named is never contacted. Matching is exact, as it is for the org-level `issuers` list, so a trailing slash makes a different issuer.

Entries are validated at startup against the same `IsValidIssuer` rule, so an entry no issuer could ever equal is reported rather than silently contributing nothing to a list whose whole purpose is to be exhaustive.

## Scope, and who this is for

This is an operator-level knob, so it protects only deployments that set it. That is a real limitation and I want to be upfront about it: a deployment serving organizations that bring their own issuers, or one relying on `issuer_patterns`, cannot use it and should leave it unset.

I think it still earns its place, because the deployments that *can* use it are the common case — a fixed, known set of identity providers, often just one — and for those it removes the pre-authentication network reach entirely rather than narrowing it. It also gives a GHES deployment a supported way to name an issuer that is not reachable from the public internet.

## Testing

- `issuerAllowed` is table-tested: empty list, listed, unlisted, trailing slash, and a suffix of a listed issuer.
- `TestExchangeChecksAllowlistBeforeDiscovery` points the issuer at a stub server and asserts it receives **zero** requests, pinning the ordering that gives the check its value.
- `TestExchangeWithoutAllowlistPerformsDiscovery` is the counterpart: with no allowlist the issuer is still contacted, so an existing deployment is unaffected.
- Two `BaseConfig` cases cover a valid list and a rejected entry.

`go build ./...`, `go vet` and `go test ./pkg/...` are clean.

## Note

`NewSecurityTokenServiceServer` gains a parameter. Happy to move the options to a struct instead if you would rather not grow the signature again.

### DIFF
--- a/pkg/octosts/allowed_issuers_test.go
+++ b/pkg/octosts/allowed_issuers_test.go
@@ -1,0 +1,142 @@
+// Copyright 2026 Chainguard, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package octosts
+
+import (
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"net/http"
+	"net/http/httptest"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	v1 "chainguard.dev/sdk/proto/platform/oidc/v1"
+	"github.com/go-jose/go-jose/v4"
+	josejwt "github.com/go-jose/go-jose/v4/jwt"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+func TestIssuerAllowed(t *testing.T) {
+	const gh = "https://token.actions.githubusercontent.com"
+
+	for _, tc := range []struct {
+		name    string
+		allowed []string
+		issuer  string
+		want    bool
+	}{{
+		name:   "empty allowlist permits any issuer",
+		issuer: "https://example.com",
+		want:   true,
+	}, {
+		name:    "listed issuer is permitted",
+		allowed: []string{gh, "https://accounts.example.com"},
+		issuer:  "https://accounts.example.com",
+		want:    true,
+	}, {
+		name:    "unlisted issuer is refused",
+		allowed: []string{gh},
+		issuer:  "https://example.com",
+		want:    false,
+	}, {
+		// Matching is exact, matching the org-level issuers list, where a
+		// trailing slash likewise makes a different issuer.
+		name:    "trailing slash is a different issuer",
+		allowed: []string{gh},
+		issuer:  gh + "/",
+		want:    false,
+	}, {
+		name:    "suffix of a listed issuer is refused",
+		allowed: []string{gh},
+		issuer:  gh + ".example.com",
+		want:    false,
+	}} {
+		t.Run(tc.name, func(t *testing.T) {
+			s := &sts{allowedIssuers: tc.allowed}
+			if got := s.issuerAllowed(tc.issuer); got != tc.want {
+				t.Errorf("issuerAllowed(%q) = %t, wanted %t", tc.issuer, got, tc.want)
+			}
+		})
+	}
+}
+
+// TestExchangeChecksAllowlistBeforeDiscovery pins the ordering that gives the
+// allowlist its value: a refused issuer must never be contacted. The issuer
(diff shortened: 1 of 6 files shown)


pkg/octosts/allowed_issuers_test.go
```diff
@@ -0,0 +1,142 @@
...
+	ctx := metadata.NewIncomingContext(context.Background(), metadata.MD{
+		"authorization": []string{"Bearer " + signedTokenForIssuer(t, srv.URL)},
+	})
+
+	s := &sts{allowedIssuers: []string{"https://token.actions.githubusercontent.com"}}
+	_, err := s.Exchange(ctx, &v1.ExchangeRequest{Identity: "foo", Scope: "org/repo"})
+	if err == nil {
+		t.Fatal("Exchange() succeeded, wanted a refusal")
+	}
+	if got := status.Code(err); got != codes.InvalidArgument {
```
jmeridth, line 89: :robot: (low, non-blocking)

Nice that this test pins the security-relevant ordering. One way to make it self-contained: `codes.InvalidArgument` is also what the preceding "invalid issuer format" check returns, so in isolation this assertion would pass vacuously if the stub URL failed `IsValidIssuer`. It is genuinely closed by `TestExchangeWithoutAllowlistPerformsDiscovery` (which proves the stub URL does reach discovery without an allowlist), but asserting the distinct message here would let it stand on its own:

```go
if got, want := status.Convert(err).Message(), "issuer is not allowed"; got != want {
    t.Errorf("Exchange() message = %q, wanted %q", got, want)
}
```

jmeridth:
:robot: (summary)

Reviewed with two models (Opus 4.8, Fable 5), verification-first; both fetched the PR head, built it, and ran the tests. Clean — no blocking findings. Nice work, and it closes the series neatly: the allowlist is the policy layer (pre-network), #1598 the egress guard, #1599 the resource bound — they compose without overlap.

Verified: the gate is complete (`provider.Get` has exactly one caller, `Exchange`, and the check sits after `IsValidIssuer` and strictly before it), empty/unset stays allow-all so existing deployments are unaffected, a bad or empty entry fails startup rather than silently disabling the gate, exact-match matches the org-level `issuers` semantics, and the issuer is only logged after `IsValidIssuer` (no log-injection vector).

A few low, non-blocking notes: entries are not whitespace-trimmed, so `"a, b"` fails startup (fail-fast is fine — worth a one-line README note that entries can not have surrounding spaces); an explicitly-empty value reads as allow-all (already documented); plus the inline test nit. On the constructor-signature question you raised: growing it is fine here, but if it grows again an options struct would read better — not needed for this PR.

jmeridth:

epol-mongodb:
I would very much like to see this functionality available soon, as this security concern has recently come up for us as well. Given the PR feedback is fairly simple, I opened https://github.com/lcasassa/app/pull/1 against the fork created by @lcasassa addressing the feedback as given.

PR #1616, opened 2026-08-28
feat(webhook): emit trust policy audit events as CloudEvents

TomPeas:
## Summary

The webhook already knows when a trust policy is created, changed, or removed — it validates them and writes a check run — but that knowledge was never published anywhere. The only CloudEvent this app emits is `dev.octo-sts.exchange`, from the token exchange path.

This adds a second stream, `dev.octo-sts.policy`, emitted from `handlePush` when trust policy changes land **on the default branch**. Together with the exchange stream it gives a detection signal for trust policy registration: who registered which policy, when, and whether it was valid — correlatable against who later used it to mint a token.

One event is emitted **per policy change**, not per push. That keeps the CloudEvent `subject` (`org/repo/policy`) at the granularity you actually alert on, keeps BigQuery rows queryable without `UNNEST`, and matches the grain of the existing exchange stream so the two join naturally.

## Why?

Implements a detection signal for trust policy changes allowing them to be mapped together with the exchanges in a observability system. They are best effort and GitHub history is the complete source of truth for auditting trust policy changes.

## Example events

A push adding one policy and deleting another (`change_count: 2`):

```json
{
  "org": "octo-sts",
  "repo": "app",
  "ref": "refs/heads/main",
  "commit": "9f2c1d7e5a3b4c6d8e0f1a2b3c4d5e6f7a8b9c0d",
  "before": "1a2b3c4d5e6f7a8b9c0d9f2c1d7e5a3b4c6d8e0f",
  "installation_id": 12345678,
  "actor": "octocat",
  "actor_id": 583231,
  "pusher": "octocat",
  "forced": false,
  "change": {
    "path": ".github/chainguard/deploy.sts.yaml",
    "policy": "deploy",
    "action": "created"
  },
  "detection": "commits",
  "change_index": 0,
  "change_count": 2,
  "valid": true,
  "time": "2026-08-28T13:45:12Z"
}
```

The deletion from that same push. Note `"valid": null` — a deleted path doesn't exist at the head SHA, so it is never read, and reporting it as valid would be a lie:

```json
{
  "org": "octo-sts",
  "repo": "app",
  "ref": "refs/heads/main",
  "commit": "9f2c1d7e5a3b4c6d8e0f1a2b3c4d5e6f7a8b9c0d",
  "before": "1a2b3c4d5e6f7a8b9c0d9f2c1d7e5a3b4c6d8e0f",
  "installation_id": 12345678,
  "actor": "octocat",
  "actor_id": 583231,
  "pusher": "octocat",
  "forced": false,
  "change": {
    "path": ".github/chainguard/legacy.sts.yaml",
    "policy": "legacy",
    "action": "deleted"
  },
  "detection": "commits",
  "change_index": 1,
  "change_count": 2,
  "valid": null,
  "time": "2026-08-28T13:45:12Z"
}
```

An invalid policy merged to the default branch — the failing check run is now also a queryable signal:

```json
{
  "org": "octo-sts",
  "repo": "app",
  "ref": "refs/heads/main",
  "commit": "9f2c1d7e5a3b4c6d8e0f1a2b3c4d5e6f7a8b9c0d",
  "before": "1a2b3c4d5e6f7a8b9c0d9f2c1d7e5a3b4c6d8e0f",
  "installation_id": 12345678,
  "actor": "octocat",
  "actor_id": 583231,
  "pusher": "octocat",
  "forced": false,
  "change": {
    "path": ".github/chainguard/broken.sts.yaml",
    "policy": "broken",
    "action": "updated"
  },
  "detection": "commits",
  "change_index": 0,
  "change_count": 1,
  "valid": false,
  "error": ".github/chainguard/broken.sts.yaml: error unmarshaling JSON: while decoding JSON: json: unknown field \"subject_patern\"",
  "time": "2026-08-28T13:45:12Z"
}
```

A rewind force-push that brings a deleted policy back. The push carries *no commits at all*, so a commit-derived stream would say nothing here:

```json
{
  "org": "octo-sts",
  "repo": "app",
  "ref": "refs/heads/main",
  "commit": "3c4d5e6f7a8b9c0d1a2b3c4d5e6f7a8b9c0d1a2b",
  "before": "9f2c1d7e5a3b4c6d8e0f1a2b3c4d5e6f7a8b9c0d",
  "installation_id": 12345678,
  "actor": "mallory",
  "actor_id": 424242,
  "pusher": "mallory",
  "forced": true,
  "change": {
    "path": ".github/chainguard/legacy.sts.yaml",
    "policy": "legacy",
    "action": "created"
  },
  "detection": "snapshot",
  "change_index": 0,
  "change_count": 1,
  "valid": true,
  "time": "2026-09-03T09:12:44Z"
}
```

A forced push whose pre-push SHA is no longer reachable. There is no `change`, because the marker is a statement about the push rather than about any one policy:

```json
{
  "org": "octo-sts",
  "repo": "app",
  "ref": "refs/heads/main",
  "commit": "3c4d5e6f7a8b9c0d1a2b3c4d5e6f7a8b9c0d1a2b",
  "before": "9f2c1d7e5a3b4c6d8e0f1a2b3c4d5e6f7a8b9c0d",
  "installation_id": 12345678,
  "actor": "mallory",
  "actor_id": 424242,
  "pusher": "mallory",
  "forced": true,
  "detection": "degraded",
  "detection_error": "resolving 9f2c1d7e...: GET .../commits/9f2c1d7e...: 404 No commit found for SHA",
  "change_index": 0,
  "change_count": 0,
  "valid": null,
  "time": "2026-09-03T09:12:44Z"
}
```

CloudEvent envelope for all of the above:

| attribute | value |
| --- | --- |
| `type` | `dev.octo-sts.policy` |
| `source` | `https://github.com/octo-sts/app` |
| `subject` | `octo-sts/app/deploy` |
| `githubinstallationid` | `12345678` |
| `githubcommit` | `9f2c1d7e...` |

A marker event's `subject` is `octo-sts/app` — no policy segment, since it names none. Note that `change` is omitted rather than nulled when absent, so a consumer must handle its absence.

## Forced pushes

A commit-derived stream has a suppression primitive built into it, and it is available to exactly the actor the stream exists to watch. A policy is deleted through a recorded push (`deleted`), then someone with default-branch push access rewinds `main` to before that deletion. The policy is live again — `octosts` reads trust policies at default-branch HEAD — but a pure rewind (`git reset --hard <ancestor>; push -f`) carries `forced: true` with an **empty** `commits` array, so nothing would be emitted and the stream's last word on that policy would remain `deleted`. The truncated-push path does not save it either: a rewind three-dot-compares as empty.

The same root cause makes the gap wider than the rewind case. Detection derived from the commits a push *carries* cannot describe what a forced push *does to the branch*. A rebase force-push has a non-empty `commits` array and passes every gate, yet still omits the path of a deletion commit dropped from the rewritten history.

So a forced push to the default branch is now detected by comparing **state**, not commits: `policyChangesFromSnapshot` lists `.github/chainguard` at `before` and at `after` and diffs by blob SHA. Restorations, silent drops, and content changes all fall out of that comparison regardless of what history the push carries.

`policySnapshot` resolves the ref before listing its contents, which matters more than it looks. GitHub answers `404` both for "this ref has no policy directory" and for "this SHA no longer exists", and conflating the two would report a repository's entire policy set as deleted. Resolving first means the `404` can only mean the former.

When `before` genuinely cannot be resolved — GitHub is free to collect it after a rewind — detection **degrades rather than fails**: validation falls back to the commit list, and a push-level marker carrying `detection: "degraded"` and the reason is emitted. Silence is precisely what suppression looks like, so the gap has to be published rather than merely logged. Consumers should treat any `detection` other than `snapshot` or `commits` as reduced assurance.

## Emission is asynchronous

GitHub allows a webhook around ten seconds to respond and treats an overrun as a failed delivery, which it then redelivers. Sending inline meant a slow or unavailable ingress could stall the response, delay the policy validation the same request performs, and — through that redelivery — duplicate the very events the sink was failing to accept.

`PolicyEmitter` (`pkg/webhook/emitter.go`) moves delivery off the request path: a bounded queue, four workers, a 30s per-event budget covering retries, and a `Shutdown` that drains with a deadline. `Enqueue` never blocks. Under sustained sink failure the choice is between dropping events and stalling webhook responses, and stalling would take policy validation down with it — so drops are counted and logged loudly instead.

## Design notes

**`valid` is a tri-state (`*bool`).** `null` means "never read", and is deliberately distinct from `false`. It covers deletions, a validation pass aborted by GitHub rate limiting, and a push that failed before validation ran (`push_error` explains the last case). Being rate-limited is specifically *not* recorded as a verdict — it says nothing about whether a policy is valid, and a consumer must not read it as a policy failure.

**Verdicts are per file.** `validatePolicies` now returns `(map[string]error, error)`: the map carries each file's own verdict, the error is the unchanged aggregate that drives the check run. Previously one broken policy in a push would have tainted every other policy's record and copied its error text into them.

**Change detection and validation share one source.** `handlePush` builds `[]PolicyChange` from the push payload, `CompareCommits`, or a snapshot, then derives the files to validate from it. GitHub truncates the payload's commit list at 20, so deriving the reported view separately would have silently under-reported large pushes — exactly the shape of push someone would use to slip a policy in. The compare path is also the only one that sees renames, which GitHub reports as a single entry plus `PreviousFilename`; those are recorded as a create plus a delete. There are now three detection paths, and the one used is published on every event as `detection`, so a consumer can weigh a record by how it was derived rather than having to assume.

**`change_index` / `change_count`** restore the grouping that per-change emission loses. Delivery is best-effort per event, so a consumer grouping by `commit` can compare the rows it holds against `change_count` to detect a partial delivery — and can still spot a bulk policy rewrite.

**Emission is not gated on `METRICS_ENABLED`,** unlike the exchange stream. Turning off metrics must not silently disable a security signal. It is gated on `EVENT_INGRESS_URI` alone.

## Behaviour change to check runs

**A policy added (or modified) and then deleted across commits in the same push no longer produces a failing check run.** Sharing one source between reporting and validation means `handlePush` now considers a commit's `Removed` list, which the old `filesFromPushEvent` ignored entirely. Previously the path was collected from `Added`/`Modified` and never removed again, so it was sent for validation, `GetContents` 404'd at the head SHA, and the push failed its check run for a policy that no longer exists. The last action for a path now wins, so it resolves to `deleted` and is excluded. This is the cross-commit case of the rule the old code already applied within a single diff — its doc comment gave exactly this reasoning for skipping removals.

I diffed the old and new file selection across add-only, add-then-modify, delete-then-re-add, modify-then-delete, add-then-delete, and mixed pushes. Every other case selects the same set. `add then modify` previously yielded a duplicate entry, which `validatePolicies` already collapsed via `sets.List`, so the validated set is unchanged there too.

**Forced pushes to the default branch are now validated against live state.** Because the snapshot replaces the commit list as the source for *both* reporting and validation, a rewind that restores a broken policy will now fail a check run where it previously produced none — and a forced push whose net effect on policies is nil no longer produces one. Both follow from validating what is actually live rather than what the push happened to carry, which is the same principle the check run already served.

The snapshot runs on forced default-branch pushes **whether or not `EVENT_INGRESS_URI` is set**. Gating it on the emitter would have made validation behave differently depending on whether eventing was configured, which is a worse trap than the extra API calls — and forced pushes to a default branch are rare.

The `CompareCommits` path selects exactly what it did before: only `removed` entries are dropped, and a rename's `PreviousFilename` was never validated previously either. `handleSHA`'s check run logic, the aggregate `multierror` and its ordering, the rate-limit short-circuit, and `handlePullRequest`/`handleCheckSuite` are all untouched.

## Testing

`go test ./... -race` passes. New coverage in `pkg/webhook`:

- `TestPushEmitsPolicyEvents` — end to end: one event per change, valid → `true`, unreadable → `false` with its own error, deletion → `null`, plus `change_index`/`change_count` integrity.
- `TestPushNonDefaultBranchEmitsNothing` — the stream stays scoped to policies that are actually live.
- `TestForcedPushRewindEmitsRestoredPolicy` — a rewind with an empty commit list restores a policy; asserts `detection: snapshot` and an authoritative `valid`.
- `TestForcedPushDegradedEmitsMarker` — an unreachable `before` emits a marker with no `change` and a populated `detection_error`.
- `TestPolicyChangesFromSnapshot` — create/update/delete/no-op diffing, ref creation, and the unreachable-`before` case that must error rather than read as an empty tree.
- `TestPolicyEmitterDoesNotBlockOnAStalledSink`, `TestPolicyEmitterShutdownDrains`, `TestPolicyEmitterShutdownRespectsDeadline` — non-blocking enqueue with counted drops, drain on shutdown, and giving up on a wedged sink.
- `TestValidatePoliciesPerFileVerdicts` — one broken policy must not taint its siblings.
- `TestPolicyChangesFromCompare` — added/modified/removed/renamed handling and non-policy filtering.
- `TestPathsToValidateExcludesDeletions`, `TestPolicyName` (including the `.trusted-token-issuers` sentinel that keeps subjects well-formed for the org allowlist).

`filesFromPushEvent` was folded into `pathsToValidate(policyChangesFromPushEvent(...))`; its test is now `TestPathsToValidateFromPushEvent`, with the deduplication case updated to expect a single entry.


pkg/webhook/webhook.go
```diff
@@ -347,30 +540,99 @@ func (e *Validator) handlePush(ctx context.Context, event *github.PushEvent) (*g
...
+
+	// Validation and auditing share one view of the push so the two can't
+	// disagree about which policies it touched.
+	files := pathsToValidate(changes)
+
+	// validation is populated by handleSHA below; the deferred closure reads it
+	// after the fact, so a push that returns early leaves it nil and every
+	// policy is reported with an unknown verdict rather than a clean one.
+	var validation map[string]error
+	if e.CEClient != nil && len(changes) > 0 && event.GetRef() == "refs/heads/"+event.GetRepo().GetDefaultBranch() {
```
jmeridth: :robot: (high)

The gate reading the signed payload and scoping to the default branch is right. But `len(changes) > 0` opens a suppression gap that matters for something framed as a security control: a **rewind force-push emits nothing**. A pure rewind (`git reset --hard <ancestor>; push -f`) carries `forced: true` but an empty `commits` array, so `policyChangesFromPushEvent` yields zero changes and the defer never registers.

Concretely: a malicious policy is deleted via an audited push (recorded `deleted`), then someone with default-branch push access force-pushes `main` back to before the deletion — the policy is live again and the stream records nothing; its last word is `deleted`. The >=20 path does not save it either (a rewind three-dot-compares as empty).

Cross-stream correlation (an exchange event naming a policy with no registration event) can still catch it, but the registration audit is suppressible by exactly the actor it is meant to watch. Given the framing I would want a mitigation before this ships as an audit trail — e.g. a forced-push snapshot / reversed-compare (diff before->after even when `commits` is empty) that emits the resurfaced policies, or at minimum a distinct "forced push, change detection skipped" tombstone event so the gap is visible rather than silent.

pkg/webhook/webhook.go
```diff
@@ -347,30 +540,99 @@ func (e *Validator) handlePush(ctx context.Context, event *github.PushEvent) (*g
...
+		ce.SetSubject(event.GetRepo().GetFullName() + "/" + change.Policy)
+		ce.SetTime(pe.Time)
+		ce.SetExtension("githubinstallationid", strconv.FormatInt(installationID, 10))
+		ce.SetExtension("githubcommit", sha)
+		if err := ce.SetData(cloudevents.ApplicationJSON, pe); err != nil {
+			log.Errorf("failed to set cloudevents data: %v", err)
+			continue
+		}
+		rctx := cloudevents.ContextWithRetriesExponentialBackoff(context.WithoutCancel(ctx), retryDelay, maxRetry)
+		if res := e.CEClient.Send(rctx, ce); cloudevents.IsUndelivered(res) || cloudevents.IsNACK(res) {
```
jmeridth: :robot: (medium)

Emission runs synchronously in the defer — one blocking `Send` per change with up to `maxRetry` backoff, on `context.WithoutCancel(ctx)` so it is not cancelled when the request ends — and it delays the HTTP response until every event is sent. A bulk policy push against a slow or down ingress stacks N x (retries + backoff) and can exceed GitHubs ~10s webhook delivery timeout, which GitHub then treats as a failed delivery and redelivers — re-emitting the whole push (duplicate events + more load on the sink). Worth bounding total emit time, or emitting on a detached goroutine / queue so a slow sink cannot stall or time out the webhook.

pkg/webhook/webhook.go
```diff
@@ -41,6 +44,167 @@ const (
...
+	ChangeCount int `json:"change_count"`
+	// Valid reports whether THIS policy parsed at Commit, and is nil when the
+	// policy was never read: deletions (the path no longer exists at Commit),
+	// a validation pass aborted by GitHub rate limiting, or a push that failed
+	// before validation ran. A nil Valid must not be read as "valid" — it means
+	// unknown, and PushError usually explains why.
+	Valid *bool `json:"valid"`
+	// Error is this policy's own validation failure, not the push's aggregate.
+	Error string `json:"error,omitempty"`
+	// PushError records a failure that prevented validation from running at
```
jmeridth: :robot: (low, non-blocking)

Two cases where `PushError` does not match this contract ("prevented validation from running at all"): (1) when `validatePolicies` succeeds but `CreateCheckRun` fails, `handleSHA` returns `(nil, results, err)` and the emit sets both authoritative per-file `Valid` **and** `push_error` — a consumer would read "validation never ran" while `valid` is actually correct; (2) on a rate-limit abort the never-read files get `valid: null` with an **empty** `push_error`, despite the sibling doc saying it "usually explains why." Either scope `PushError` to the true pre-validation-abort case, or relax the doc.

jmeridth:
:robot: (summary)

Reviewed with four models (Opus 4.8, Sonnet 5, Haiku 4.5, Fable 5), verification-first — full control-flow trace, scratch tests for the change-detection orderings, and a cross-check against the exchange stream. Implementation quality is high and the mechanics are exactly as described: the deferred-emit control flow captures the right `validation`/`err` on every return path, the per-file verdict map + the added `continue`s are load-bearing and correct (a failure never falls through to `valid`), check-run behavior is invariant, the tri-state `valid` maps correctly in all five cases (clean / invalid / deletion / rate-limited / push-failed), the default-branch gate reads the signed payload, and rename / dedup / ordering all hold. (The nil-`Pusher` panic a scan might flag is a non-issue — go-github getters are nil-safe.)

The thing to weigh is the "security control" framing versus what is actually built — a best-effort mirror of the push diffs GitHub volunteers:

- **High (inline):** a rewind force-push emits nothing — a suppression primitive available to the exact actor the audit is meant to watch. Worth a forced-push snapshot / tombstone path before it ships as an audit trail.
- **Medium:** this rewrites `handlePush` and does not carry #1591s zero-SHA fix, so a new-ref / import push (`before` = zero) with >=20 commits still 500s on `CompareCommits` and now also produces no audit event — and #1591 is still open, so the two need reconciling. Emit is synchronous and can approach GitHubs webhook timeout under a slow sink (inline). And the headline behavior-change (add-then-delete no longer fails a check run) ships with no regression test for that exact scenario.
- **Low:** the `PushError` / `valid:null` doc mismatches (inline); `policyChangesFromCompare` records GitHubs `unchanged` status as an update; `policyName` degenerates to an empty subject for a literal `.sts.yaml`.
- **Pre-existing (not this PR):** the `CompareCommits` file list is capped at 300 and unpaginated, so a >300-file push can hide a policy from both validation and audit.

Net: genuinely useful detection signal and clean code — I would frame it as that rather than an integrity-guaranteed audit trail until the forced-push and change-detection-failure gaps are closed or documented. Nice work on the tri-state and per-file verdicts in particular.

pkg/webhook/webhook.go
TomPeas: :robot: (summary)

Fixed in `aeb29e0`.

Agreed on the gap, with one refinement to the diagnosis: `len(changes) > 0` is the symptom rather than the cause. Removing that gate on its own changes nothing, because on a rewind `changes` is already empty — the real problem is that the stream was derived from **commit diffs** at all. That makes the gap wider than described: a *rebase* force-push carries a non-empty `commits` array and so passes every gate, but still omits the path of any dropped deletion commit. So it isn't only pure rewinds that go unrecorded.

The fix is therefore a **state** comparison rather than a commit comparison. On a forced push to the default branch we now list `.github/chainguard` at `before` and at `after` and diff by blob SHA (`policySnapshot` / `policyChangesFromSnapshot`), which catches the resurfaced policy regardless of what the commit list says. Events carry a new `detection` field (`commits` / `compare` / `snapshot` / `degraded`) so a consumer can tell which path produced a record.

Two things worth calling out:

- The contents API returns 404 both for "no policy directory at this ref" and "that SHA no longer exists". Conflating them would report a repo's entire policy set as deleted, so `policySnapshot` resolves the ref via `GetCommit` first; only then is a 404 from `GetContents` safely read as "empty".
- GitHub can GC `before` after a rewind, so we degrade rather than fail: validation falls back to the commit list and we emit a push-level `detection: "degraded"` marker carrying `detection_error`. That is your tombstone suggestion — the gap is visible in the stream rather than silent — but it now only applies to the case where the history genuinely isn't there to compare against.

One deliberate consequence, flagged in the PR body: the snapshot runs on forced default-branch pushes regardless of whether `EVENT_INGRESS_URI` is set, so validation doesn't quietly behave differently based on eventing config. A rewind restoring a broken policy now fails a check run where it previously produced none, and a forced push with net-zero policy effect no longer produces one.

Covered by `TestPolicyChangesFromSnapshot`, `TestForcedPushRewindEmitsRestoredPolicy` and `TestForcedPushDegradedEmitsMarker`.

On the "before it ships as an audit trail" framing — two responses. First, the PR description now describes this as a **detection signal** rather than an integrity-guaranteed audit trail, which is the honest description of what a mirror of GitHub's volunteered push diffs can be. Second, on sequencing: nothing is emitted in production yet. `EVENT_INGRESS_URI` isn't wired into `modules/app/webhook.tf` and `dev.octo-sts.policy` isn't registered in the broker's `types` map in either env. Those are follow-up PRs, so there's a natural gate between this merging and it becoming a live signal.

pkg/webhook/webhook.go
TomPeas: :robot: (summary)

Fixed in `aeb29e0`, and the two other Medium items from your summary are addressed below.

**Synchronous emit.** You're right that this was the wrong thing to do inline, and the redelivery consequence you describe is the part that matters: an overrun doesn't just delay a response, it makes GitHub re-send the push and re-emit the whole thing, which adds load to a sink that was already the reason we were slow.

Emission has moved behind a `PolicyEmitter` (now in its own file, `pkg/webhook/emitter.go`): a bounded queue of 2048 feeding 4 workers, with a 30s per-event send timeout. `Enqueue` is non-blocking, so the handler's cost is now a channel send and the HTTP response no longer waits on the ingress at all.

The design decision worth flagging is what happens when the queue fills: we **drop, counted and logged**, rather than blocking. Blocking would mean a failing ingress takes policy *validation* down with it, and check runs are the load-bearing function here — the event stream is the secondary signal. Dropping keeps the failure in the subsystem that failed. Queue depth is also a useful sink-health signal, because the buffer sits near-empty in the healthy case.

`cmd/webhook/main.go` drains on shutdown: `Shutdown` is deferred with a 15s deadline on `context.WithoutCancel(ctx)`, so in-flight events get a chance to land after `ListenAndServe` returns rather than being discarded with the process.

Covered by `TestPolicyEmitterDoesNotBlockOnAStalledSink`, `TestPolicyEmitterShutdownDrains` and `TestPolicyEmitterShutdownRespectsDeadline`. One knock-on: because delivery is now per-event and asynchronous, a partially-delivered push is possible, so events carry `change_index` / `change_count` to make that detectable.

**Add-then-delete regression test.** Fair hit — the headline behaviour change shipped untested. Added `TestPushAddThenDeleteCreatesNoCheckRun` in `efc1767`, which drives a real push through the handler and asserts **zero** check runs are created for a policy that doesn't exist at the head SHA, while the deletion is still emitted with `valid: null` and no `push_error`. Also added three cross-commit cases to `TestPathsToValidateFromPushEvent`: add→delete, modify→delete, and delete→re-add (which must still be validated, because the policy *is* live at head).

**#1591 reconciliation.** Deliberately not carried here, since it's still open and unapproved. I've structured the detection `switch` in `handlePush` so their zero-SHA arm slots in as one additional `case` before `default` rather than landing in conflicting territory; whichever merges second should be a small rebase. Happy to invert that if you'd rather #1616 absorb it.

pkg/webhook/webhook.go
TomPeas: :robot: (summary)

Fixed in `efc1767`, along with the other two Low items from your summary.

**`PushError` / `valid: null` mismatch.** Both cases confirmed. The underlying problem was that `handleSHA` returned `(map[string]error, error)` where the second value conflated "validation was cut short" with "something else went wrong afterwards", and the emit path couldn't tell them apart.

Scoped it, as you suggested, by making that distinction explicit in the type. `handleSHA` now returns a `validationOutcome{verdicts, incomplete}`, where `incomplete` is set **only** by a failure that limited how much of the push could be validated:

- **Case (1), `CreateCheckRun` fails after a successful pass** — the per-file verdicts are authoritative, so they're reported and `push_error` is no longer set alongside them. The check-run failure is still returned to the handler and still drives the 5xx; it just isn't presented to a consumer as doubt about a verdict we actually have.
- **Case (2), rate-limit abort** — `incomplete` is now populated, so the never-read files report `valid: null` **with** a populated `push_error` explaining why the answer is missing. Previously they carried the null with nothing to explain it, which is the worse of the two failures: silently indistinguishable from a deletion.

The doc comments were part of the problem, so both were rewritten. `PushError` now says it describes what happened to the *push* and explicitly does not imply the policy went unvalidated; `Valid` notes that deletions are the other reason it's null.

Worth noting the rate limit deliberately does **not** produce a 5xx — a redelivery would just amplify the rate limit we're already hitting. That's why the incompleteness has to travel on the events rather than being signalled by the response code, and it's recorded as a comment on the struct so it doesn't get "tidied up" later. Covered by `TestRateLimitedPushReportsNoVerdict`.

**`unchanged` recorded as an update.** Confirmed — it fell into the `default` arm. The status switch is now explicit: `modified`/`changed` → update, `unchanged` → skipped. An *unrecognised* status still reports as an update, deliberately: over-reporting is the safer failure for a signal meant to catch policy movement, and a future GitHub status value silently vanishing from the stream would be worse than a spurious record. It now logs a warning so the gap is visible rather than inferred.

**`policyName` empty subject.** Confirmed, and it's specifically because Go's `filepath.Match` lets `*` match the empty string, so `.github/chainguard/.sts.yaml` matches the glob octo-sts actually validates. Fixed with a fallback to the filename, plus a test case pinning it.

**`CompareCommits` 300-file cap.** Agreed it's real and agreed it's pre-existing. Tracking it separately rather than growing this PR — it affects validation as much as the event stream, so it deserves its own change and its own tests.

TomPeas:
Thanks for the review @jmeridth sorry it took so long to get round to adrdessing the comments. Agree with this not being phrased as hardline, its a best effort event log intended for easy viewing in observability platforms. I have made the emitting more robust using queues and workers to preserve as much as possible but its still fallible. I have omitted any changes from #1591 as they're not yet approved, this should be easy enough to rebase onto and vice versa.

pkg/webhook/webhook.go
jmeridth: :robot: (low, non-blocking)

The async `PolicyEmitter` resolves this — non-blocking enqueue, counted drops, per-event budget, and the handler no longer waits on the ingress. Drop-on-full keeping validation up rather than the stream is the right tradeoff. Two low hardening notes for a fast-follow:

- `Enqueue` after `Shutdown` panics: `select { case queue <- ev: default }` still panics on a send to a *closed* channel (the `default` only covers full-but-open — I checked with a standalone repro). Unreachable in todays `main` since `Shutdown` only runs after `ListenAndServe` returns, but a `done`-channel select or a closed flag would harden it against any future graceful-shutdown change.
- The drain defer only fires on the `log.Panic(ListenAndServe())` unwind. Cloud Run / k8s terminate with SIGTERM, which `signal.NotifyContext(..., os.Interrupt)` does not catch, so defers do not run and the drain does not happen on the common shutdown path — the "drain in-flight on shutdown" benefit mostly will not fire. `SIGTERM` + a graceful `srv.Shutdown` would make it real. (Minor: the 30s per-event budget also exceeds the 15s shutdown deadline, so one in-flight send can pin a worker past it.)

pkg/webhook/webhook.go
jmeridth: :robot: (low, non-blocking)

Resolved — the `validationOutcome{verdicts, incomplete}` split plus the "read Valid, not PushError" contract is a clean way to disambiguate, and the rewritten docs are clear. One small note: your reply above says `push_error` is "no longer set" on the CreateCheckRun-failure path, but the code (and the new doc) keep it set there, with the "a non-nil Valid alongside a PushError is an authoritative verdict" clarification — which I think is the better design, so no change needed, just flagging that the reply and the code describe it differently. The `unchanged` -> skipped and `policyName` fallback fixes both check out.

pkg/webhook/webhook.go
jmeridth: :robot: (low, non-blocking)

Confirmed resolved for the resolvable-`before` case — the blob-SHA snapshot is the right call, the `GetCommit`-before-`GetContents` 404 disambiguation is correct, and the rewind now emits a `snapshot`/`created`/`valid:true` event for the resurfaced policy (verified + pinned by test). Nice sharpening on the rebase-drops-a-deletion variant.

One residual, and I am explicitly not blocking on it: when `before` has been GC'd, the degraded path falls back to `policyChangesFromPushEvent`, which for a pure rewind (empty commits) yields no changes — so no check run and no per-policy event for the resurfaced policy fire, only the content-free `degraded` marker. It is visible now (the important improvement over round 1) and needs GitHub to have GC'd the SHA, so it is a narrow residual rather than the original hole. Since only `before` failed to resolve, the `after` snapshot is still available — validating/listing that on degrade (instead of the commit list) would keep the check-run gate and a per-policy record even when the diff cannot be computed. Fine as a fast-follow given prod emission is gated off.

jmeridth:
:robot: (summary)

Re-reviewed round 2 with four models (Opus 4.8, Sonnet 5, Haiku 4.5, Fable 5), verification-first (full trace + `go test -race`, all green). Thorough response — every substantive finding from round 1 is resolved with pinning tests:

- **HIGH force-push** -> resolved via the blob-SHA `before`/`after` snapshot; the rewind now records the resurfaced policy. One narrow residual in the degraded / GC'd-`before` fallback, noted inline — non-blocking, fast-follow.
- **MEDIUM sync-emit** -> resolved via the async `PolicyEmitter`; a failing sink can no longer take validation down. Two low emitter-lifecycle hardening notes inline (Enqueue-after-Shutdown panic; SIGTERM drain) for a fast-follow.
- **MEDIUM untested behavior-change** -> regression tests added.
- **LOWs** (PushError/valid, unchanged-status, policyName) -> resolved.

I re-verified the previously-good properties survived the refactor + main-merge (default-branch gate reads the signed payload, tri-state valid, rename = create+delete, dedup/ordering, check-run invariance, and #1597's missing-policy-dir fix), and confirmed emission is gated off in prod (no `EVENT_INGRESS_URI` in `webhook.tf`, `dev.octo-sts.policy` not registered in the broker) — so nothing goes live on merge.

Net: in good shape. Nothing I'd block on. The degraded-fallback residual and the two emitter LOWs are reasonable fast-follows. Thanks for the careful rework — the `detection` method field and the visible `degraded` marker in particular are a real improvement over "silence = suppression."

jmeridth:
:robot: (summary)

The 09-07 update is a clean merge of `main` (CI workflows, `go.mod`/`go.sum`, and `modules/app` version bumps only) with no changes under `pkg/webhook` or `cmd/webhook`, so my round-2 review still stands as-is.

Nothing blocking from me (emission stays gated off in prod). The three round-2 fast-follows are the outstanding items whenever you pick this back up:

- degraded fallback: on a GC'd `before`, validate/list the `after` snapshot instead of the commit list, so a rewind still produces a check run and a per-policy record rather than only the content-free marker;
- `Enqueue` after `Shutdown`: a send on the closed queue panics even with the `default` case, so a guard (done channel or closed flag) would harden it against any future graceful-shutdown change;
- SIGTERM drain: the shutdown defer only fires on the `log.Panic(ListenAndServe())` unwind, not on the SIGTERM that Cloud Run/k8s send, so the drain does not run on the common shutdown path.

No rush, just flagging they are still open.

TomPeas:
> 🤖 (summary)
> 
> The 09-07 update is a clean merge of `main` (CI workflows, `go.mod`/`go.sum`, and `modules/app` version bumps only) with no changes under `pkg/webhook` or `cmd/webhook`, so my round-2 review still stands as-is.
> 
> Nothing blocking from me (emission stays gated off in prod). The three round-2 fast-follows are the outstanding items whenever you pick this back up:
> 
>     * degraded fallback: on a GC'd `before`, validate/list the `after` snapshot instead of the commit list, so a rewind still produces a check run and a per-policy record rather than only the content-free marker;
> 
>     * `Enqueue` after `Shutdown`: a send on the closed queue panics even with the `default` case, so a guard (done channel or closed flag) would harden it against any future graceful-shutdown change;
> 
>     * SIGTERM drain: the shutdown defer only fires on the `log.Panic(ListenAndServe())` unwind, not on the SIGTERM that Cloud Run/k8s send, so the drain does not run on the common shutdown path.
> 
> 
> No rush, just flagging they are still open.

Thanks @jmeridth have addressed all the above in this commit https://github.com/octo-sts/app/pull/1616/commits/8370dccb178300fbb0b53c647768e26a8cfd5ea6

jmeridth:
:robot: (summary)
Re-reviewed the delta (`8370dccb`) since my last pass. The graceful-shutdown concurrency model is correct: `Enqueue` holds `RLock` across the check-and-send while `Shutdown` takes the write `Lock` before `close(queue)`, so a send can't interleave with the close (no panic); workers derive from `p.sendCtx` with `defer drainCancel()`, so the 6s emitter budget actually cuts a wedged send; shutdown drains the server (3s) before closing the emitter (6s), and `context.WithoutCancel` rebuilds the deadlines. Covered by a `-race` test and an in-flight-cancellation test. The bundled `PolicyPresent` forced-push change strengthens the audit path (validates a resurfaced policy when only `before` is GC'd) rather than weakening it; no regression to the blob-SHA snapshot or async emit. No high/medium.

- Low, non-blocking: server (3s) + emitter (6s) drain = 9s against Cloud Run's 10s SIGKILL, leaving ~1s for tracer flush + `kmsClient.Close` (`main.go:33-34`); fast in practice but almost no headroom.
- Nit: the `PolicyPresent` change is unrelated to graceful shutdown but rides the same commit/title.

Approve, merge as-is.

jmeridth (approved):

PR #1623, opened 2026-09-01
feat: add dynamodb support

lee2sh:


pkg/stickystore/dynamodb/dynamodb.go
```diff
@@ -0,0 +1,121 @@
...
+	Scope          string    `dynamodbav:"scope"`
+	Identity       string    `dynamodbav:"identity"`
+	Subject        string    `dynamodbav:"subject"`
+	CreatedAt      time.Time `dynamodbav:"created_at"`
+	ExpireAt       time.Time `dynamodbav:"expire_at,unixtime"`
+}
+
+// New creates a DynamoDB-backed sticky store. Items are keyed by the "id"
+// string partition key in the given table, with TTL managed via the
+// expire_at attribute (the table must have TTL enabled on expire_at).
```
jmeridth, line 45: :robot: (medium)
This doc comment is the only place the mandatory `expire_at` TTL is stated. Firestore ships `modules/app/firestore.tf` creating the DB, IAM, and TTL field policy; this PR adds no equivalent, and the three new `OCTOSTS_STICKY_STORE_DYNAMODB_*` env vars are undocumented (README:299-306 documents the Firestore ones). If an operator points at a table without TTL enabled on `expire_at`, nothing errors and stale `(scope, identity) → installation` mappings are pinned forever, the exact eviction this store exists to provide. Please document the env vars + the TTL requirement, and add Terraform parity if AWS is an in-repo deploy target.

jmeridth:
:robot: (summary)
Should we do this: yes. The backend is correct and well-tested: strongly-consistent reads (`ConsistentRead: true`), injection-safe keys, and a subtle `expire_at < :new` condition that refuses to resurrect a TTL-deleted item. Faked-client coverage is thorough.

One medium (inline). Approve once addressed.

Low (non-blocking): region/endpoint isn't configurable, `newDynamodb` (`factory.go:44`) uses `LoadDefaultConfig` only, so unset `AWS_REGION` → client builds but every call fails at runtime, silently falling back to round-robin; consider `OCTOSTS_STICKY_STORE_DYNAMODB_REGION`/`_ENDPOINT` (the endpoint also enables a dynamodb-local test). Expired-but-not-yet-deleted items can still be served as hits (matches Firestore, parity not regression).

lee2sh:
> 🤖 (summary) Should we do this: yes. The backend is correct and well-tested: strongly-consistent reads (`ConsistentRead: true`), injection-safe keys, and a subtle `expire_at < :new` condition that refuses to resurrect a TTL-deleted item. Faked-client coverage is thorough.
> 
> One medium (inline). Approve once addressed.
> 
> Low (non-blocking): region/endpoint isn't configurable, `newDynamodb` (`factory.go:44`) uses `LoadDefaultConfig` only, so unset `AWS_REGION` → client builds but every call fails at runtime, silently falling back to round-robin; consider `OCTOSTS_STICKY_STORE_DYNAMODB_REGION`/`_ENDPOINT` (the endpoint also enables a dynamodb-local test). Expired-but-not-yet-deleted items can still be served as hits (matches Firestore, parity not regression).

will look at this once my other PR gets merged. Thanks!

Issue #1600, opened 2026-08-21
provider.Get has no negative cache / single-flight for stalled or failing issuers


jmeridth:
Surfaced during the review of #1599 (which bounds a single request's discovery cost to 3 attempts / ~18s).

### Problem

`provider.Get` (`pkg/provider/provider.go`) caches only successful providers. There is no negative cache and no single-flight around `newProviderWithRetry`. Because the issuer comes from an unverified bearer token, an unauthenticated caller chooses the discovery target.

With #1599 merged, one inbound request against a stalled or failing issuer is bounded (3 attempts, ~18s, 3 outbound requests). But N concurrent inbound requests naming the same bad issuer each independently pay that full cost — the aggregate outbound load and held-connection count scale linearly with inbound volume. #1599 fixes per-request amplification; this is the remaining aggregate surface.

### Proposal

- Single-flight discovery per issuer (e.g. `golang.org/x/sync/singleflight`) so concurrent requests for the same issuer collapse to one in-flight discovery.
- Short-TTL negative cache for recently-failed issuers so a stalling/failing target is not re-probed on every request during the failure window.

### Notes

- Mirrors the single-flight mitigation already applied to the org-allowlist lookup in #1560.
- Bounds should stay conservative: a negative cache must expire quickly enough that a genuinely recovering issuer is retried.

jmeridth:
Both proposals shipped, so closing as done:

- Single-flight discovery per issuer: `provider.Get` now collapses concurrent calls for the same issuer through a `singleflight.Group` (`discoveryFlight`), landed in #1629.
- Short-TTL negative cache for failing issuers: a bounded 2Q LRU (`negativeCache`, cap 100) with a 5s TTL remembers recently-failed discoveries, landed in #1630.

Concurrent inbound requests naming the same bad issuer collapse to one in-flight discovery, and a stalling or failing target is not re-probed on every request during the window. The 5s TTL keeps a genuinely recovering issuer from being punished for long.

PR #1629, opened 2026-09-02
perf(provider): single-flight concurrent OIDC discovery per issuer

jflacker:
Part of #1600; the negative cache half is a separate PR, #1630, so each can be reviewed and reverted independently.

Concurrent Get calls for the same issuer now collapse into one discovery using singleflight, so N callers naming a slow or stalled issuer no longer each pay for a full discovery. An earlier version of this PR tied the shared discovery to whichever caller happened to trigger it, so one caller's short timeout could fail discovery for other callers with a longer or no deadline at all. That is fixed: the shared discovery now runs on its own fixed timeout (20 seconds) independent of any single caller, and each caller still returns promptly if its own context expires while waiting.

One tradeoff is intentional and remains. If every caller waiting on an issuer gives up early, the shared discovery keeps running in the background up to that 20 second cap anyway, since it is no longer tied to any specific caller's cancellation. This cost is bounded per issuer and shared across concurrent callers, so it does not grow with how many callers gave up.

New tests cover the singleflight collapse, the context isolation fix, and a caller still respecting its own deadline while waiting. The full pkg/provider suite passes under race across multiple runs, and golangci-lint, go vet, and gofmt are all clean.

### DIFF
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -28,6 +28,14 @@ import (
 //   - Chainguard: needs around 2KiB
 const MaximumResponseSize = 100 * 1024 // 100KiB
 
+// discoveryTimeout bounds a single shared discovery, independent of any one
+// caller's own context. Discovery is single-flighted across concurrent
+// callers (see discoveryFlight below), so it must not be tied to any single
+// caller's deadline: an unrelated caller's short timeout must not abort
+// discovery for other callers sharing the same issuer, and a genuinely
+// stalled issuer must not hold the shared call open forever either.
+const discoveryTimeout = 20 * time.Second
+
 var (
 	// providers is an LRU cache of recently used providers.
 	providers, _ = lru.New2Q[string, VerifierProvider](100)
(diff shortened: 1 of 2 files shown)


pkg/provider/provider.go
```diff
@@ -44,21 +58,39 @@ func Get(ctx context.Context, issuer string) (provider VerifierProvider, err err
...
-	provider, err = newProviderWithRetry(ctx, issuer)
-	if err != nil {
-		return nil, fmt.Errorf("constructing %q provider: %w", issuer, err)
+	select {
+	case res := <-ch:
+		if res.Err != nil {
+			return nil, fmt.Errorf("constructing %q provider: %w", issuer, res.Err)
+		}
+		provider = res.Val.(VerifierProvider)
+	case <-ctx.Done():
```
jmeridth, line 98: :robot: (medium)
If the only caller hits this `<-ctx.Done()` branch, the shared flight still completes and succeeds, but its result is dropped, nothing calls `providers.Add`, and the next request re-discovers from scratch. Under load with short caller deadlines against a slow-but-healthy issuer, you re-pay for discovery and throw away every success. Memoize inside the flight closure on success (call `providers.Add` there) so the result is cached regardless of which caller is still listening.

jmeridth:
:robot: (summary)
Should we do this: yes. Correctly implements the cold-start / thundering-herd mitigation from #901, and the #1598 SSRF `CheckRedirect` is preserved. `DoChan` doesn't memoize failures, so a failed flight stays retryable.

One medium (inline). Approve once addressed.

Low (non-blocking): no test asserts the background success is memoized when the leader leaves, nor that a failed flight is retryable on the next call. The isolation/collapse tests are solid.

pkg/provider/provider.go
jflacker: Fixed: `providers.Add` now runs inside the `DoChan` closure itself, right after a successful `newProviderWithRetry`, so a discovery success is cached exactly once regardless of whether the caller that triggered it is still waiting. Verified with a new test (`TestGet_LoneCallerTimeoutStillMemoizesBackgroundSuccess`): the lone caller times out, and the memoized result is confirmed present after the background flight completes. Also added `TestGet_FailedFlightIsRetryableOnNextCall` to cover the low priority item you flagged; `singleflight.Group` was already clearing completed flights correctly, this just proves it.

jmeridth:
:robot: (summary)
Re-reviewed the delta (`760e547e`, `56c78cec`) against my earlier medium. Resolved: `providers.Add` now runs inside the `DoChan` closure (`provider.go:88`), the caller-path duplicate is gone, and a failed flight stays retryable (only success memoizes; singleflight drops the key). `TestGet_LoneCallerTimeoutStillMemoizesBackgroundSuccess` reproduces the exact timeout case and `TestGet_FailedFlightIsRetryableOnNextCall` covers retryability. SSRF `CheckRedirect`/`IsValidIssuer` and fail-closed preserved. No new issues. Approve.

jmeridth (approved):

PR #1630, opened 2026-09-02
perf(provider): short-TTL negative cache for failing OIDC issuers

jflacker:
Part of #1600 (negative-cache half only; the single-flight half was #1629, now merged).

Rebased onto main and restructured per review: setNegativeCache is now called inside #1629's shared flight closure, keyed to discoveryCtx's own error rather than any caller's. discoveryCtx is independent of any caller and bounded only by discoveryTimeout, so its error (including a timeout from discoveryTimeout itself) is always a real signal about the issuer, never a caller giving up early. The context-cancellation exclusion from the pre-rebase version is no longer needed for that reason: a cancelling caller's own return path never reaches setNegativeCache at all now.

Also bounded negativeCache with the same lru.New2Q pattern as providers, since issuer is attacker-controlled.

Tested:
- TestGet_NegativeCacheAvoidsRepeatedProbesForFailingIssuer / TestGet_NegativeCacheExpiresAndRetriesRecoveredIssuer / TestGet_NegativeCacheIsBoundedAgainstManyDistinctFailingIssuers carried over and passing
- New: TestGet_CallerCancellationNeverReachesNegativeCache, reproducing the original poisoning scenario end to end and confirming a fresh caller succeeds
- New: TestGet_SharedFlightTimeoutIsNegativeCached, proving a genuinely stalled issuer (shared-flight timeout, not caller cancellation) is now correctly negative-cached, which was the residual gap flagged for this rebase
- Full pkg/provider suite passes under -race (2x), golangci-lint/go vet/gofmt clean

https://claude.ai/code/session_01SNArp8JbQtboNNLWze4EUs


pkg/provider/provider.go
```diff
@@ -58,7 +86,9 @@ func Get(ctx context.Context, issuer string) (provider VerifierProvider, err err
 	// Verify the token before we trust anything about it.
 	provider, err = newProviderWithRetry(ctx, issuer)
 	if err != nil {
-		return nil, fmt.Errorf("constructing %q provider: %w", issuer, err)
+		wrapped := fmt.Errorf("constructing %q provider: %w", issuer, err)
+		setNegativeCache(issuer, wrapped)
```
jmeridth: :robot: (medium)
`newProviderWithRetry` ran on the caller's ctx, so on cancellation `err` is `context.Canceled`/`DeadlineExceeded`, and caching it here poisons a healthy issuer for all clients for the full TTL. A client (or attacker) can name a healthy issuer, disconnect mid-discovery, and every other client is served that cached error for 5s, repeatable → sustained DoS. Don't cache context errors:

```go
if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
    return nil, wrapped
}
```

before `setNegativeCache` (ideally only negative-cache permanent-class failures).

jmeridth:
:robot: (summary)
Should we do this: yes, but land it after #1629. Bounded (`lru.New2Q`, 100), fail-closed, ordered positive→negative→discovery; it can only deny, never bypass auth or SSRF checks.

One medium (inline). Also: this and #1629 rewrite the identical block (`provider.go` ~47-62), so they will conflict and the resolution is security-relevant. Please land #1629 first and rebase this on top, placing `setNegativeCache` so it caches only the shared flight's discovery error (background ctx), not any caller path. Done that way, #1629's background-ctx decoupling neutralizes the poisoning vector for free.

Low (non-blocking): `negativeCacheTTL` is a package-level mutable `var` only for test override (safe today, no `t.Parallel`); Info-level log on every cache hit (`provider.go:71`) could flood, consider Debug; no test that context/transient errors are not cached.

pkg/provider/provider.go
jflacker: Fixed: `setNegativeCache` is now skipped for `context.Canceled` and `context.DeadlineExceeded`. Verified with a new test (`TestGet_CancelledCallerDoesNotPoisonHealthyIssuerForOthers`) that reproduces the exact scenario you described (disconnect mid discovery against a healthy issuer) and confirms a following caller is no longer denied.

On sequencing: agreed on landing #1629 first and rebasing this on top with `setNegativeCache` keyed to the shared flight's error rather than any caller path. One thing worth flagging in the meantime: on this branch alone, `newProviderWithRetry` still runs on the caller's own context, so a genuinely stalled issuer's timeout is indistinguishable from a caller giving up early, meaning stalled issuers aren't negative cached here either until the rebase. Documented that directly in the code so it's not an implicit gap.

jmeridth:
:robot: (summary)
Re-reviewed the delta (`4231077b`, `02934db6`) against my earlier medium. Resolved: context errors are excluded from `setNegativeCache` (`provider.go:104`), and `TestGet_CancelledCallerDoesNotPoisonHealthyIssuerForOthers` reproduces the poisoning and asserts a fresh caller succeeds. Bounded cache, positive-first ordering, and TTL expiry all intact; fail-closed preserved. Approve, with two coordination notes:

- Land after #1629 (both rewrite the same block; this branch isn't rebased on it yet).
- Low, non-blocking (at rebase): once on top of #1629, a shared-flight timeout arrives via `res.Err` as `DeadlineExceeded` and *should* be cacheable, but the current blanket `DeadlineExceeded` exclusion would wrongly suppress it. Move `setNegativeCache` into the `res.Err != nil` branch (as your `02934db6` comment anticipates) and drop the caller-ctx exclusion there.

pkg/provider/provider.go
jflacker: Rebased on top of #1629 and restructured as suggested: `setNegativeCache` now lives inside the shared flight closure, keyed to `discoveryCtx`'s own error rather than any caller's. Added `TestGet_SharedFlightTimeoutIsNegativeCached` to prove the specific case you called out (a genuine shared-flight timeout is now correctly cacheable) and `TestGet_CallerCancellationNeverReachesNegativeCache` to confirm caller cancellation still can't poison it, now structurally rather than via an explicit error-type check. Full suite green under -race.

jmeridth (approved):
:robot: (summary)
Re-reviewed the rebase onto #1629 (`55abb4d5`). The composition is now correct: `setNegativeCache` lives inside the shared-flight closure on the caller-independent `discoveryCtx` error path (`provider.go:80`), so a stalling-issuer `DeadlineExceeded` is cached while caller cancellation returns via `<-ctx.Done()` without ever writing the cache — the poisoning vector is structurally closed, not just guarded, and the blanket `DeadlineExceeded` exclusion is correctly gone. `TestGet_CallerCancellationNeverReachesNegativeCache` and `TestGet_SharedFlightTimeoutIsNegativeCached` cover both halves. Ordering, success memoization, SSRF `CheckRedirect`, fail-closed, and the bounded LRU are all intact. My earlier medium is resolved. Approve.

PR #1635, opened 2026-09-04
App specific routing

lee2sh:
Trust policies can now pin which GH app mints their tokens: 

```
# ci.sts.yaml
issuer: https://token.actions.githubusercontent.com
subject: repo:my-org/my-repo:ref:refs/heads/main
permissions:
  contents: read
app: reviewer-bot
# app_pattern: ci-*
```

In multi-app deployments, exchanges are normally load-balanced across the pool. However, sometimes a workload requires a specific gh app -- different permission grants or a dedicated rate limit

This PR lets trust policies pin the App that mints their tokens: apps get an optional `app_name` in the server config, and a policy selects one with `app` or `app_pattern` in its sts.yaml

Policies without a pin behave exactly as before (covered by tests), and broken pins fail loudly rather than silently re-routing.

Let me know if this fits the vision of the project -- happy to change it in any way or split the PR up if that's easier to review.


pkg/octosts/trust_policy.go
```diff
@@ -43,6 +43,12 @@ type TrustPolicy struct {
 	// See https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-an-installation-access-token-for-an-app
 	Permissions github.InstallationPermissions `json:"permissions,omitempty"`
 
+	// App pins exchanges to one configured app (name or numeric app ID).
+	App string `json:"app,omitempty"`
```
jmeridth, line 47: :robot: (medium)
Adding these fields under `UnmarshalStrict` (both `octosts.go` and `webhook.go:303-325`) means a policy using `app:` is rejected as an unknown field by any binary predating this PR. If policies adopt `app:` before both the `app` and `webhook` binaries roll out, the old webhook fails the check-run on a valid file and/or the old exchange binary returns `NotFound`. It's fail-closed but a foot-gun, please document the locked-step deploy ordering (ship both binaries, then allow the fields).

pkg/octosts/octosts.go
```diff
@@ -294,12 +328,191 @@ func (s *sts) getExchangeInstall(ctx context.Context, pool *ghinstall.OrgPool, o
...
+	insts, enumErr := pool.M.GetAll(ctx, owner)
+	candidates := eligibleInstalls(insts, eligible)
+	fresh := false
+	if len(candidates) == 0 {
+		if enumErr != nil {
+			return nil, 0, enumErr
+		}
+		missErr := status.Errorf(codes.FailedPrecondition, "no installation for %q matches the trust policy app pin", owner)
+		missKey := pinMissKey(owner, eligible)
+		if _, confirmed := pinMisses.Get(missKey); confirmed {
```
jmeridth, line 389: :robot: (medium)
This confirmed-miss short-circuit returns before any fresh enumeration for the full 5-min TTL, which defeats the `GetAllFresh` recovery it sits on. Deploy `app: ci-bot`, run one exchange (confirms the miss), then install `ci-bot` seconds later, and exchanges keep returning `FailedPrecondition` for up to 5 minutes with no way to force a refresh. Consider a shorter miss-TTL, or don't cache the very first confirmed miss so install-then-retry recovers on the next request.

pkg/octosts/octosts.go
```diff
@@ -294,12 +328,191 @@ func (s *sts) getExchangeInstall(ctx context.Context, pool *ghinstall.OrgPool, o
...
+				}
+				if !present && !confirmed {
+					pinMisses.Add(absKey, struct{}{})
+				}
+			}
+			clog.FromContext(ctx).Infof("sticky install %d not in pinned app set for %s, reassigning", cachedID, owner)
+		}
+	}
+
+	pick := candidates[routekey.Index(scope, identity, subject, len(candidates))]
```
jmeridth, line 463: :robot: (medium)
Pinned routing picks purely by `routekey.Index(...) % len(candidates)`, bypassing the capacity-aware `pickByQuota` the default path uses. For a single-app pin this is moot, but for `app_pattern` spanning several apps it can deterministically land every matching caller on a rate-limit-exhausted installation while an eligible sibling has headroom. Reuse the quota picker over the eligible set, or document that pins opt out of capacity-aware balancing.

jmeridth:
:robot: (summary)
Should we do this: yes. The `app`/`app_pattern` pin is genuinely fail-closed: unresolvable pin, no-match, or pinned-app-not-installed all return `FailedPrecondition` and abort rather than re-route. No auth bypass or cross-org leak (per-org pools only enumerate their own apps). Concurrency (singleflight `GetAllFresh`, detached recording, partial-enumeration) is careful and well-tested. No high.

Three mediums (inline) to address or document before merge.

Low (non-blocking): the webhook compiles the policy but has no `AppSet`, so a pin to a nonexistent app passes validation (green check) yet fails at exchange, worth a doc note; no user-facing docs for the new fields (20 files changed, none are docs); sticky-reassign across a newly-narrowed pin can break same-app check-run ownership (#901), one-line warning worth adding.

pkg/octosts/octosts.go
lee2sh: Hi @jmeridth! Thanks for the review! I applied the capacity aware routing when `app_pattern` is in use. LMK what you think

pkg/octosts/octosts.go
lee2sh: Hey @jmeridth! thanks for the review. I've shortened the TTL to 1 minute. LMK what you think.

pkg/octosts/trust_policy.go
lee2sh: Hey @jmeridth! thanks for the review. I've updated the README.

jmeridth:
:robot: (summary)
Re-reviewed at `875d891` (Opus 4.8 + Fable 5). All three mediums are resolved: locked-step deploy ordering is documented (README) with fail-closed handling on old binaries; the pin-miss `pinConfirmTTL` is now 1 min and only caches after an err-free single-flighted `GetAllFresh`; and pinned routing uses the shared `PickByQuota` for non-checks:write pins while checks:write pins hash over a canonically-sorted candidate set (`9ee4717c`) with a stable-set confirmation guard, preserving same-app check-run ownership. No auth bypass or cross-org leak; fail-closed on unresolvable pins. `go build`/`go test -race ./pkg/octosts/...` pass locally at head. Nice work.

One blocker: it's conflicting with main, and the conflict is in `trust_policy.go`/`webhook.go` — the files the GHSA-mwqh-2vg8-rhj3 fix rewrote. Please resolve by adopting main's shared `compileAnchored` (`pkg/octosts/pattern.go`) for `app_pattern` and main's `parseAndCompile` in the webhook, rather than keeping this PR's inline copies — the two anchoring paths must not diverge.

Non-blocking follow-ups: `app_pattern` should call the shared `compileAnchored` (currently inline); a checks:write pin with no sticky store can't recover from an App uninstall until restart (worth an inline note).

jmeridth (approved):
:robot: (summary)
Approving. The rebase onto main is done and the one blocker is resolved: `app_pattern` now uses the shared `compileAnchored` (`pattern.go`) and the webhook uses `parseAndCompile`, so there is no duplicate anchoring path. The three earlier mediums stay resolved (locked-step deploy ordering documented, pin-miss `pinConfirmTTL` at 1 min, quota-aware `PickByQuota` for non-checks:write pins with a canonically-sorted deterministic pick for checks:write). The `small nonblocking fixes` commit also documents the checks:write-pin uninstall-recovery limitation. Checks are green at head.

PR #1643, opened 2026-09-06
fix: fully anchor trust policy regex patterns

ajayk:
regexp.Compile("^" + p + "$") leaves patterns with a top-level alternation unanchored on one side: "|" binds looser than the anchors, so "main|develop" compiles to (^main)|(develop$) and matches the subject "main-attacker" as a prefix (or "attacker-develop" as a suffix). A policy like

  subject_pattern: repo:org/app:ref:refs/heads/main|repo:org/app:ref:refs/heads/develop

therefore also matches repo:org/app:ref:refs/heads/main-anything, letting a caller who can mint a token for a prefix-extended subject (e.g. by creating a branch named main-x) satisfy a policy meant for main only. Claim patterns leak in the suffix direction too: a branch named x/refs/tags/v1 yields ref refs/heads/x/refs/tags/v1, which matches the refs/tags/v1 alternative.

Compile all four pattern kinds (issuer_pattern, subject_pattern, audience_pattern, claim_pattern) as "^(?:" + p + ")$", mirroring the org trusted-issuer allowlist compilation in org_trusted_issuers.go, which already does this and documents why. The pattern is compiled alone first so an unbalanced ")" is rejected instead of escaping the wrapper group.

This tightens matching only for policies that were already leaking: exact and properly grouped patterns behave identically.


jmeridth (approved):
:robot: (summary)
4-model review (Opus 4.8, Sonnet 5, Haiku 4.5, Fable 5), all approve. Verified the bypass is real and closed at all four pattern sites via `compileAnchored`: `"^"+p+"$"` parses a top-level `|` as `(^A)|(B$)` (e.g. `main|develop` matched `main-attacker`), and the new `^(?:p)$` rejects the leak while the standalone-compile-first step provably prevents the `foo)|(` group-escape (an unbalanced pattern is rejected before the wrap can re-open it). Backward-compat is strictly narrowing (no new `Compile()` failures; ReDoS-free under RE2), and this is the only remaining unanchored identity-gating path besides the already-hardened `org_trusted_issuers.go`. It also closes an inline `(?m)` flag leak reachable via claim values. Nice fix.

I also audited every octo-sts trust policy in our mono, stereo, and infra repos: every alternation there is already wrapped in a top-level group (`subject_pattern: "(A|B|C)"`), which was never vulnerable and matches identically under `^(?:...)$` — no bare top-level `|` and no inline flags anywhere, so this merge changes matching for zero deployed policies.

Non-blocking follow-ups: (1) add `issuer_pattern`/`audience_pattern` alternation regression tests to lock all four fields; (2) have `OrgTrustedIssuers.Compile` call the shared `compileAnchored` so the two paths can't drift; (3) optional: make the webhook check-run `Compile()` trust policies so a bad pattern is caught at PR time instead of exchange time.

jmeridth:
@ajayk I'll let this sit for a day in case you want to add the nits to this PR. Otherwise I'll merge and we can do a follow-up PR. Your call. Thank you for your contribution.

ajayk:
> @ajayk I'll let this sit for a day in case you want to add the nits to this PR. Otherwise I'll merge and we can do a follow-up PR. Your call. Thank you for your contribution.

I will add a follow up , lets merge this

jmeridth:
Thanks @ajayk — this landed the core anchoring fix and shipped in [v0.10.0](https://github.com/octo-sts/app/releases/tag/v0.10.0), now covered by the published advisory [GHSA-mwqh-2vg8-rhj3](https://github.com/octo-sts/app/security/advisories/GHSA-mwqh-2vg8-rhj3).

Nothing outstanding on your end. The follow-ups I'd flagged (routing the org allowlist through the shared `compileAnchored`, compiling policies in the webhook check-run, and per-field alternation tests) were all folded into the coordinated advisory fix by @egibs. Appreciate the contribution.

PR #1671, opened 2026-09-16
close secret provider at webhook shutdown

phelian:
## What/Why

`NewSecretProvider` can create a GCP Secret Manager client backed by a gRPC connection, but `SecretProvider` exposes no cleanup method and `cmd/webhook` never closes it. The connection therefore remains open after its startup-only work has completed.

This makes `SecretProvider` implement `io.Closer`, matching the existing `kms.KMS` lifecycle contract. The GCP provider delegates cleanup to its Secret Manager client; AWS and Azure retain no-op cleanup because their SDK clients do not expose a close operation. The webhook scopes provider cleanup to startup secret loading, releasing the client as soon as the secrets are in memory.

## Proof it works

- `go test -race ./...`
- `go test ./...` with `CGO_ENABLED=0`
- `golangci-lint run ./...`
- `gofmt` and `goimports`
- `go mod tidy` produced no module changes
- `actionlint`
- `misspell` and boilerplate checks

The new unit tests verify delegated cleanup, close-error propagation, no-op cleanup for providers without a closable client, and the real GCP constructor-to-closer wiring.

## Risk + AI role

Low: this changes shutdown cleanup only. GCP closes its existing Secret Manager gRPC client; AWS and Azure behavior remains a no-op.

Diagnosed and authored with AI assistance using `github-copilot/gpt-5.6-sol`; reviewed by the author.

## Review focus

Whether embedding `io.Closer` in `SecretProvider`, parallel to `kms.KMS`, is the preferred lifecycle contract.


cmd/webhook/main.go
```diff
@@ -115,6 +115,7 @@ func main() {
 		if err != nil {
 			log.Panicf("could not create secret provider: %v", err)
 		}
+		defer secretsProvider.Close() //nolint:errcheck // released at process shutdown
```
cpanato: :robot: (medium)

`defer` here is function-scoped, so this fires at the end of `main()` — but `secretsProvider` is already dead after the loop below: the values land in `webhookSecrets` and the `Validator` only ever sees plain `[][]byte`. So the GCP gRPC connection is still held for the full process lifetime, which is the thing the description calls out as the problem.

This is where the `kms.KMS` parallel stops holding, I think — `kmsClient` really is process-lifetime since it signs a JWT on every delivery, so `defer` at the end of `main` is right for it. This one is startup-only.

Scoping it to the fetch would release the client when it stops being useful:

```go
		func() {
			secretsProvider, err := secrets.NewSecretProvider(ctx, baseCfg.KMSProvider)
			if err != nil {
				log.Panicf("could not create secret provider: %v", err)
			}
			// The provider is only needed to read the secrets below, so release it
			// here rather than holding the client open for the process lifetime.
			defer secretsProvider.Close() //nolint:errcheck // secrets are already in memory
			for name := range strings.SplitSeq(webhookConfig.WebhookSecret, ",") {
				name = strings.TrimSpace(name)
				val, err := secretsProvider.GetSecret(ctx, name)
				if err != nil {
					log.Panicf("error fetching webhook secret %s: %v", name, err)
				}
				webhookSecrets = append(webhookSecrets, val)
			}
		}()
```

I checked this builds, is `gofmt`-clean, and that appending to `webhookSecrets` through the closure still works. The `log.Panicf` paths still run the deferred `Close` on the way out.

Happy either way, but if you'd rather keep the current placement, could the description be reworded? As written it promises the connection stops being held until process termination, and it still is.

pkg/secrets/secrets_test.go
```diff
@@ -31,6 +42,21 @@ func TestSecretProvider_GetSecretReturnsErrOnFakeProvider(t *testing.T) {
...
+	sp := &secretProvider{closer: closer}
+
+	err := sp.Close()
+
+	assert.True(t, closer.closed)
+	assert.ErrorIs(t, err, wantErr)
+}
+
+func TestSecretProvider_CloseWithoutCloser(t *testing.T) {
+	assert.NoError(t, (&secretProvider{}).Close())
```
cpanato, line 58: :robot: (low, non-blocking)

Both of these build `&secretProvider{}` directly, so `NewSecretProvider` never runs — I ran a coverage profile and the block with `sp.closer = client` in the GCP branch has count 0. Dropping that one line would leave the suite fully green while quietly turning `Close()` back into a no-op in the only environment where it does anything.

Worth pinning down? `gcpSM.NewClient` dials lazily, so it succeeds offline with `GOOGLE_APPLICATION_CREDENTIALS` pointing at a throwaway service-account JSON — enough to assert `NewSecretProvider(ctx, "gcp")` returns a provider whose `Close()` is actually wired through.

cpanato (commented):
:robot: (summary)

Reviewed with two models (Opus + Fable) at `a29239b`.

The shape is right — embedding `io.Closer` to match `kms.KMS` is the established contract in this repo, so to your review-focus question: yes, that's the preferred one. Worth knowing it makes `SecretProvider` a breaking change for any out-of-repo implementer, but that's the same tradeoff already taken for `kms.KMS`, and in-repo blast radius is nil.

Both models traced and cleared the things worth worrying about: all `log.Panic*` exit paths still run the defer, there's no `os.Exit` anywhere in the tree, the AWS/AKV no-ops are correct (neither SDK client exposes `Close` at the pinned versions), no constructor path creates a client and then errors, and there's no `Close`/`GetSecret` race.

One substantive comment inline about where the `Close` lands, plus a non-blocking note on test coverage for the GCP wiring.

cmd/webhook/main.go
phelian: Agreed. Fixed in 481ff63 by scoping provider creation, secret loading, and deferred cleanup inside a closure. The GCP client is now released immediately after the secrets have been copied into `webhookSecrets`, including on the `log.Panicf` paths.

pkg/secrets/secrets_test.go
phelian: Added in 481ff63. `TestNewSecretProvider_GCPClose` uses a temporary offline credential file, calls the real `NewSecretProvider(ctx, GCP)` path, asserts the GCP client is wired as the closer, and closes it successfully. Constructor coverage is now 87.5%; removing `sp.closer = client` breaks this test.

cmd/webhook/main.go
cpanato: :robot: Confirmed at `481ff63` — the provider is now closed before the mux and server exist, so this actually releases the client instead of parking it for the process lifetime. I checked the shadowing risk the closure introduces: the outer `err` is last read at the `ghtransport.New` call, and everything after is a fresh `:=`, so nothing downstream sees the inner one. Thanks for taking it.

pkg/secrets/secrets_test.go
cpanato: :robot: Verified — I mutation-tested it: removing `sp.closer = client` fails `TestNewSecretProvider_GCPClose` at the `assert.Same`. Coverage matches your 87.5% exactly. Also checked it's CI-safe, since that was my worry with a real constructor in a test — `GOOGLE_APPLICATION_CREDENTIALS` outranks ambient ADC and the metadata server, and it passes in 0.00s with the network blackholed. Nice approach.

cpanato (approved):
:robot: (summary)

Re-reviewed at `481ff63` with two models (Opus + Fable). Both round-one findings are addressed and I verified each independently: the IIFE releases the GCP client before the server starts, and the new GCP test is load-bearing — deleting `sp.closer = client` fails it.

`go test -race ./...` passes clean. Neither model found anything new in the delta.

LGTM.

PR #1672, opened 2026-09-16
fix(octosts): fail closed when an org pool has no AppIDs

jmeridth:
## What/Why

`eligibleApps` treated a nil `OrgPool.AppIDs` as "membership unknown, allow every configured app", so an `app`/`app_pattern` pin resolved against the global app set instead of the owner's pool. App pinning is an org-isolation control, so an absent pool map must not widen it. A future router-construction path that forgets to set `AppIDs` would silently reopen cross-org pin resolution through Go's zero value.

Not reachable today (`buildPool` always sets `AppIDs`), so this is a hardening follow-up to #1635, flagged by that PR's 4-model review (Sonnet 5 + Haiku 4.5 converged on it). Both branches now fail closed: a nil-map lookup returns `false`, so dropping the `AppIDs != nil &&` / `AppIDs == nil ||` special-casing makes an absent pool resolve to no eligible app.

## Proof it works

`go build ./...`, `go vet`, and `go test ./pkg/octosts/... ./pkg/ghinstall/...` pass on go 1.27.1. The two `eligible_apps` tests that asserted fail-open now assert `FailedPrecondition`; the shared `poolOf` helper and the inline quota-pool / anchored-pattern fixtures that had leaned on nil `AppIDs` now set them explicitly.

## Risk + AI role

low -- one access-control default flips from fail-open to fail-closed, with no production behavior change (`AppIDs` is always set today). AI-assisted (Claude Opus 4.8); human-directed.

## Review focus

That fail-closed is the correct default for absent pool membership, and that no production path constructs an `OrgPool` with nil `AppIDs` expecting "allow all".


pkg/octosts/octosts_test.go
```diff
@@ -1232,7 +1232,10 @@ func TestExtractUserAgent(t *testing.T) {
 }
 
 func poolOf(m ghinstall.Manager) *ghinstall.OrgPool {
-	return &ghinstall.OrgPool{M: m, AppCount: 3}
+	// These tests configure apps 101 (ci-a), 102 (ci-b), and 201 (deploy), so
+	// mark them all as pool members. eligibleApps fails closed on a nil AppIDs
+	// map, and production always sets it, so the pool must set it here too.
```
cpanato: :robot: (low, non-blocking) Nice catch making this fixture explicit — without it the `hidden` pools at 1635/1644/1698 would have started failing closed. One thing: the comment lists 101/102/201 but the map also has 103, and 103 isn't spurious — it's `ci-c` in `pin_order_test.go:23` and `pin_snapshot_test.go:113`, both of which share `poolOf` and pin `ci-.*`. I removed 103 to check, and `TestChecksWritePinIgnoresConfiguredAppOrder` fails all three subtests with "routes selected 2 eligible installations, want all 3". Worth naming it so nobody trims the map to match the comment:

```suggestion
	// Shared by the pin tests in this package, which configure apps 101
	// (ci-a), 102 (ci-b), 103 (ci-c, from pin_order_test.go and
	// pin_snapshot_test.go), and 201 (deploy), so mark them all as pool
	// members. eligibleApps fails closed on a nil AppIDs map, and production
	// always sets it, so the pool must set it here too.
```

pkg/octosts/octosts_test.go
```diff
@@ -1232,7 +1232,10 @@ func TestExtractUserAgent(t *testing.T) {
...
+	return &ghinstall.OrgPool{M: m, AppCount: 3, AppIDs: map[int64]bool{101: true, 102: true, 103: true, 201: true}}
```
cpanato, line 1240: :robot: (low, non-blocking) Side effect of populating `AppIDs` here: `"pinned app not installed"` (line 1399) pins `ghost`=301 against this pool, and 301 isn't a member, so it now returns early from `eligibleApps` with "is not configured for" instead of reaching the installation-enumeration path its name describes. I confirmed with call counters — `freshCalls=0` now, vs `1` at base where nil `AppIDs` skipped the membership check and `pinMissErr` did the rejecting. Both return `FailedPrecondition`, so the assertion can't tell the difference. Not a coverage gap (`"concurrent pin misses share one confirmation walk"` and `"cancelled caller still records the detached confirmation"` still exercise `pinMissErr` for a named pin via `deploy`/201), just a subtest that no longer matches its name. A local pool with `AppIDs: {301: true}` in that subtest would restore it.

cpanato (dismissed):
:robot: (summary) Reviewed with Opus and Fable 5. Agree with the direction: making Go's zero value the *deny* value for an access-control map is the right default, and I verified the "no production behavior change" claim — both construction sites go through `buildPool` and always set `AppIDs`, and on the legacy env path the new check is provably a no-op because the same map is aliased into the pool and `AppSet.IDs`. `go build`, `go vet`, and `go test ./pkg/octosts/... ./pkg/ghinstall/... ./cmd/...` pass on my end. Two non-blocking test-fidelity notes inline. Nit: `AppCount: 3` alongside a 4-entry `AppIDs` breaks the invariant production upholds (`cmd/app/main_test.go:78`) — harmless since nothing using `poolOf` reads `AppCount`, but `4` would be self-consistent.

pkg/octosts/octosts_test.go
jmeridth: Applied your suggestion in 82aa8de — the comment now names 103 (ci-c) and its source, so the map cannot be trimmed to match a stale comment. Thanks.

pkg/octosts/octosts_test.go
jmeridth: Fixed in 82aa8de: the subtest now uses a local pool with AppIDs {301}, so the pin reaches installation enumeration instead of returning early on the membership check. Good catch on the call counters.

egibs (approved):

PR #1684, opened 2026-09-21
fix(octosts): classify permission 403 as PermissionDenied, not rate limit

### LINKED ISSUE #1320: 403 from GitHub API misclassified as rate limit in trust policy lookup
## Problem

In \`lookupTrustPolicy\`, any 403 from GitHub is mapped to \`codes.ResourceExhausted\` with the message \`"GitHub API rate limit exceeded (403)"\`. However, GitHub returns 403 for two distinct situations:

- **Secondary rate limiting** — where \`ResourceExhausted\` is the right signal
- **Permission denied** — e.g., "resource not accessible by integration", repo access denied, insufficient installation token scope

Both land in the same bucket today. A permissions problem surfaces in logs and to callers as a rate limit, which is misleading.

## Impact

1. **Misleading error messages** make on-call debugging harder — operators chasing a "rate limit" when the real issue is a misconfigured installation or missing repo permission.
2. **Wasteful app rotation** — the outer retry loop treats \`ResourceExhausted\` as a signal to try a different app. For a genuine permissions error, no app will have access, so this burns quota on other apps unnecessarily.

## Fix

Distinguish rate-limit 403s from permission 403s. Options:

- Check for \`*github.AbuseRateLimitError\` (which wraps the 403 secondary rate limit case specifically) before falling back to the raw status code check.
- Inspect the response body or \`X-RateLimit-Remaining\` / \`Retry-After\` headers to distinguish the two.

A permission-denied 403 should probably return \`codes.PermissionDenied\` rather than \`codes.ResourceExhausted\`.

## Note

This is a pre-existing issue, not introduced by #1318. It predates the retry logic added in that PR.

jmeridth:
## What/Why

Closes #1320. A permission-denied 403 from the trust policy read was classified as a rate limit (`ResourceExhausted`), which misled operators and drove wasteful app rotation. This splits proven rate limits from bare 403s: a bare 403 now returns `PermissionDenied` and fails closed.

## Proof it works

`go build`, `go vet`, and `go test ./pkg/octosts/` pass, including `-race` on the new tests.

- `TestPolicyReadForbiddenDoesNotRotate`: a 403 on the first app returns `PermissionDenied` without rotating to a working second app.
- `TestPolicyReadRetriesOnRateLimit`: a real rate limit still rotates and recovers.
- `TestFetchTrustPolicyRawForbiddenFailsClosed`: a 403 returns `PermissionDenied`, is not retried, does not serve stale, and does not seed the negative cache.
- `TestIsProvenRateLimit` / `TestIsForbidden`: table coverage of the classifier.

## Risk + AI role

Low. Scoped to the trust policy read path. The exported lenient `IsGitHubRateLimited` is unchanged, so webhook redelivery keeps its fail-safe leniency. The strict classifier reuses the proven pattern already in `org_trusted_issuers_store.go` (now deduped into a shared `isProvenRateLimit`). AI-generated (Claude Opus 4.8) with human review.

## Review focus

1. The fail-closed choices: a bare 403 stops app rotation and refuses stale-serve. In a multi-app pool this means a second app that might have access is not tried. This matches the issue's intent (rotation is for rate-limit load spreading, not permission fallback), but it is the one behavior change worth scrutiny.
2. Branch ordering in `fetchTrustPolicyRaw`: proven-rate-limit, then permission-403, then 404, then transient.


cpanato (commented):
:robot: (summary) Multi-model review (Opus 5, Sonnet 5, Haiku 4.5, Fable 5), verified against head `945a227`.

**Should we do this? Yes.** #1320 is correctly diagnosed — `IsGitHubRateLimited`'s own doc comment says "anything that gates access needs a stricter test", and `fetchTrustPolicyRaw` was using it on exactly such a path. Right fix, right layer, and sharing the strict classifier that already existed in `org_trusted_issuers_store.go` beats a third copy.

The core claim holds up under a revert test: with `isProvenRateLimit` swapped back to `IsGitHubRateLimited` and the `isForbidden` branch deleted, both `TestPolicyReadForbiddenDoesNotRotate` and `TestFetchTrustPolicyRawForbiddenFailsClosed` fail (`code = OK` — rotation succeeded / stale served). Neither is a false pin. The fakes drive real `httptest` muxes and let go-github build the errors, so they exercise the production classification path.

Two things worth fixing before merge, both confirmed by running code — left as inline comments. Plus one comment nit.

**Also worth noting (not inline — unchanged lines):**
- `octosts.go:917-923` — the `op` preamble still says "Rate limits and 404s stop immediately"; this PR adds a third terminal class right beneath it.
- **No rotation on 403** is a real but narrow regression. The realistic heterogeneous-pool cases don't reach this branch: a repo outside the installation fails at the *mint* with 422 (`*ghinstallation.HTTPError`, not `*github.ErrorResponse`), and GitHub returns 404 for repos an installation can't see. It reduces to per-app org IP allowlists. Not a blocker.
- A persistently-misconfigured org now burns ~3 GitHub calls per exchange forever with no cooldown — still ~4x better than pre-PR rotation. A short-TTL 403 cooldown, distinct from the permanent NotFound negative cache, would close it. Out of scope here.
- `PermissionDenied` is also what legitimate policy denials return (`trust_policy.go:166-246`), so code alone no longer separates "policy rejected you" from "we can't read the policy".

**Verified clean:** build/vet/`-race` pass; the `isOrgIssuerRateLimit` dedup is semantically identical; imports all still used; `isTransient` shadows nothing (only matches >=500 or no-response); `errors.As(nil, ...)` is safe so the missing nil guard is fine; no new information disclosure; no conflict with #1682/#1683.

pkg/octosts/octosts.go
```diff
@@ -824,6 +824,35 @@ func IsGitHubRateLimited(err error) bool {
...
+		return false
+	}
+	var rateLimitErr *github.RateLimitError
+	var abuseRateLimitErr *github.AbuseRateLimitError
+	if errors.As(err, &rateLimitErr) || errors.As(err, &abuseRateLimitErr) {
+		return true
+	}
+	var errResp *github.ErrorResponse
+	return errors.As(err, &errResp) && errResp.Response != nil &&
+		errResp.Response.StatusCode == http.StatusTooManyRequests
```
cpanato: :robot: (medium) Verified with a probe against the real `fetchTrustPolicyRaw`: a 403 carrying `Retry-After: 60` but no `documentation_url` anchor comes out as `PermissionDenied` — no rotation, no stale-serve. That's the inverse of the bug being fixed.

| 403 shape | code |
|---|---|
| `X-RateLimit-Remaining: 0` | `ResourceExhausted` |
| doc_url `…#secondary-rate-limits` | `ResourceExhausted` |
| `Retry-After: 60`, JSON, no doc_url | **`PermissionDenied`** |
| `Retry-After: 60`, non-JSON body | **`PermissionDenied`** |

go-github types a secondary limit *only* on the doc_url suffix (`github.go:1754-1756`), and it matches two different spellings because GitHub has changed that string before. `Retry-After` is parsed at `github.go:1761`, inside the already-matched branch, so it never contributes to classification.

The strongest evidence is internal rather than speculative: `isMintRateLimit` in this same package (`org_trusted_issuers_store.go:172-183`) already treats `Retry-After != ""` as proof of a secondary limit, and its comment says so explicitly. This classifier contradicts its sibling.

```suggestion
	var errResp *github.ErrorResponse
	if !errors.As(err, &errResp) || errResp.Response == nil {
		return false
	}
	if errResp.Response.StatusCode == http.StatusTooManyRequests {
		return true
	}
	// Mirror isMintRateLimit: a 403 carrying Retry-After (or an exhausted
	// remaining count) is a rate limit even when go-github's body-wording
	// match failed to type it.
	return errResp.Response.StatusCode == http.StatusForbidden &&
		(errResp.Response.Header.Get("Retry-After") != "" ||
			errResp.Response.Header.Get("X-RateLimit-Remaining") == "0")
```

pkg/octosts/octosts.go
```diff
@@ -824,6 +824,35 @@ func IsGitHubRateLimited(err error) bool {
...
+}
+
+// isForbidden reports whether err is a bare GitHub 403 that isProvenRateLimit has
+// already ruled out as a rate limit: a genuine permission-denied response.
```
cpanato: :robot: (low, non-blocking) Comment nit: this phrasing implies `isForbidden` depends on `isProvenRateLimit` having run first, but that ordering isn't load-bearing. None of `ErrorResponse`, `RateLimitError`, or `AbuseRateLimitError` define `Unwrap()`, so `errors.As(rl, &errResp)` is false for the typed rate-limit errors and `isForbidden` is correct standalone — which is exactly how `TestIsForbidden` calls it.

Might be worth rewording to say it's independently safe rather than order-dependent. (Relatedly, `isProvenRateLimit`'s doc slightly overstates go-github: `*AbuseRateLimitError` only appears when the `documentation_url` suffix matches.)

pkg/octosts/octosts.go
```diff
@@ -935,6 +967,14 @@ func (s *sts) fetchTrustPolicyRaw(ctx context.Context, base *ghinstallation.Apps
...
 			return "", backoff.Permanent(status.Errorf(codes.ResourceExhausted, "GitHub API rate limit exceeded for %q", tpKey.identity))
 		}
+		// A bare 403 is permission-denied, not a rate limit. Fail closed: do not
+		// serve a stale policy (access may have been revoked) and do not rotate
+		// apps in lookupTrustPolicyWithRetry, which keys off ResourceExhausted.
+		// A permission problem is not one another app can resolve.
+		if isForbidden(err) {
+			clog.WarnContextf(ctx, "trust policy read forbidden for %q (403, not a rate limit): %v", tpKey.identity, err)
+			return "", backoff.Permanent(status.Errorf(codes.PermissionDenied, "trust policy read forbidden for %q (not a rate limit)", tpKey.identity))
+		}
```
cpanato, line 1002: :robot: (medium) This branch refuses to *serve* stale but leaves the entry in `staleTrustPolicies` (1h TTL, vs the primary cache's 5min — `octosts.go:73-74`), so the fail-closed intent isn't fully realized.

Confirmed at runtime: seed stale, then 403 → `PermissionDenied` (correct) and the stale entry is still present; then a 429 on the same key → `code=OK`, the stale policy is served **and** re-seeded into the primary cache. So a policy GitHub has already refused to let us read can still mint tokens up to an hour later if a rate limit coincides.

Strictly better than pre-PR (where the 403 itself served stale and self-renewed the TTL), but worth closing while you're here.

```suggestion
		if isForbidden(err) {
			clog.WarnContextf(ctx, "trust policy read forbidden for %q (403, not a rate limit): %v", tpKey.identity, err)
			// Access may have been revoked: drop the stale copy too, or a later
			// proven rate limit on this key would serve it (and re-seed the
			// primary cache) after GitHub already refused the read.
			staleTrustPolicies.Remove(tpKey)
			return "", backoff.Permanent(status.Errorf(codes.PermissionDenied, "trust policy read forbidden for %q (not a rate limit)", tpKey.identity))
		}
```

pkg/octosts/octosts.go
jmeridth: @cpanato Good catch, and confirmed against go-github: `AbuseRateLimitError` is only typed on the `documentation_url` anchor match (`github.go` CheckResponse), so a 403 with just `Retry-After` fell through to a bare `ErrorResponse` and came out as `PermissionDenied`. `isProvenRateLimit` now mirrors `isMintRateLimit` with a `Retry-After` / `X-RateLimit-Remaining` fallback. Addressed in https://github.com/octo-sts/app/commit/76bd287.

pkg/octosts/octosts.go
jmeridth: @cpanato Reworded, though in the opposite direction. Once the `Retry-After` fallback from the comment above lands, a rate-limited 403 also satisfies `isForbidden`, so the ordering does become load-bearing. The doc now says `isForbidden` must run after `isProvenRateLimit`. Also corrected the `AbuseRateLimitError` wording to note the anchor match. Addressed in https://github.com/octo-sts/app/commit/76bd287.

pkg/octosts/octosts.go
jmeridth: @cpanato Agreed, that undercut the fail-closed intent. The forbidden branch now removes the stale entry before returning, and a new test (`TestFetchTrustPolicyRawForbiddenThenRateLimitDoesNotResurrect`) pins that a 403 then a 429 on the same key does not serve stale. Addressed in https://github.com/octo-sts/app/commit/76bd287.

pkg/octosts/octosts.go
cpanato: :robot: (medium, resolved) Verified fixed at `76bd287`. Re-ran the same probe against the real `fetchTrustPolicyRaw`:

| 403 shape | `945a227` | `76bd287` |
|---|---|---|
| `X-RateLimit-Remaining: 0` | `ResourceExhausted` | `ResourceExhausted` |
| doc_url `…#secondary-rate-limits` | `ResourceExhausted` | `ResourceExhausted` |
| `Retry-After: 60`, no anchor | **`PermissionDenied`** | **`ResourceExhausted`** |
| `Retry-After: 60`, non-JSON body | **`PermissionDenied`** | **`ResourceExhausted`** |
| `Resource not accessible by integration` | `PermissionDenied` | `PermissionDenied` |

Also confirmed the new table cases are a real pin: deleting the header fallback fails `403 with Retry-After`, `403 with X-RateLimit-Remaining 0`, and `wrapped 403 with Retry-After`.

One thing I checked for collateral damage, since `isOrgIssuerRateLimit` delegates here: the header fallback changes the org-allowlist path too, but it's not a regression. `classifyContentsError` already gates on the rate-limit check before its own 403 case, and `record()` folds both `fetchRateLimited` and `fetchFailed` into `anyFailed`, so the stale-serve at `org_trusted_issuers_store.go:497` fires either way. The only delta is the final gRPC code when no stale exists (`Unavailable` → `ResourceExhausted`), which is more accurate.

pkg/octosts/octosts.go
cpanato: :robot: (low, resolved) You're right and I was wrong — thanks for pushing back rather than just applying it. The `Retry-After` fallback is exactly what makes the ordering load-bearing: `isForbidden` matches *any* 403, so once `isProvenRateLimit` also matches some 403s, a rate-limited one satisfies both and only the gate order separates them. Inverting the comment is the correct call.

Verified the invariant actually holds: `isForbidden` has exactly one caller (`octosts.go:998`), sitting behind the `isProvenRateLimit` gate at `:981`. The `AbuseRateLimitError` anchor-match correction reads accurately against go-github's `CheckResponse` too.

pkg/octosts/octosts.go
cpanato: :robot: (medium, resolved) Verified fixed at `76bd287`. Re-ran the resurrection probe: the 403 now evicts the stale entry, and the follow-up 429 on the same key returns `ResourceExhausted` with empty raw instead of serving the revoked policy and re-seeding the primary cache.

`TestFetchTrustPolicyRawForbiddenThenRateLimitDoesNotResurrect` is a genuine pin — removing the `staleTrustPolicies.Remove(tpKey)` line fails both it and the new stale assertion in `TestFetchTrustPolicyRawForbiddenFailsClosed`.

cpanato (approved):
:robot: (summary) Recheck at `76bd287` — all three findings verified fixed. Approving.

- **Retry-After secondary limits (medium)** — fixed. All four rate-limit 403 shapes now return `ResourceExhausted`; only a genuine permission denial returns `PermissionDenied`.
- **Stale resurrection (medium)** — fixed. The 403 evicts the stale entry; a later 429 no longer serves the revoked policy.
- **`isForbidden` doc (low)** — fixed, and reworded in the opposite direction from my suggestion, correctly. The header fallback is what makes the ordering load-bearing. Verified `isForbidden` has one caller, behind the `isProvenRateLimit` gate.

Bonus: the stale `op` preamble comment I'd only flagged in the summary was updated too.

Both new tests are real pins (confirmed by reverting each fix independently). Checked the `isOrgIssuerRateLimit` call site for collateral damage from the shared classifier change — no regression, only a more accurate gRPC code when no stale entry exists. `go build`, `go vet`, `go test ./...`, and `-race` on `pkg/octosts` all clean.

Remaining non-blockers, unchanged and fine to merge with: no app rotation on 403 (narrow — per-app org IP allowlists only, since the common cases fail at the mint with 422 or return 404), no 403 cooldown so a misconfigured org burns ~3 calls per exchange (still ~4x better than pre-PR), and `PermissionDenied` no longer separating "policy rejected you" from "can't read the policy" by code alone.

Nice work on the turnaround.

egibs (approved):

PR #1693, opened 2026-09-22
Preserve context status during installation enumeration

phelian:
## Summary

- return `Canceled` or `DeadlineExceeded` when `roundRobin` installation enumeration ends because its context is done
- preserve any installations collected before a mid-walk cancellation
- keep genuine manager failures in the returned error and the incomplete-enumeration warning even when the context ends, so a real fault (e.g. bad credentials on one app) is not hidden behind a bare `context canceled`
- avoid the incomplete-enumeration warning for plain context termination, while keeping `Unavailable` and the warning for genuine manager failures
- keep the shared pin-miss confirmation walk reporting `Unavailable` when its own `pinConfirmTimeout` expires, since no caller deadline fired

## Why

`roundRobin.enumerate` currently folds every error into `codes.Unavailable`. That includes `context.Canceled` and `context.DeadlineExceeded`, which can trigger inappropriate retries, skew availability alerting, and emit warning logs for expected caller termination.

The helper is shared by `GetAll` and `GetAllFresh`. The fix tracks the context error separately from manager failures: a manager error observed while the context is already done is attributed to the context rather than the manager, only genuine failures are collected and logged, and the returned status takes the context's code with the joined failures in its message.

`getAllFreshShared` runs the confirmation walk on a context detached from the caller and bounded by a server-side timeout. Without the second hunk, that timeout would now surface to every waiter as `DeadlineExceeded` even though none of their own deadlines expired, so it is mapped back to `Unavailable` at that one call site.

## Validation

- `go build ./cmd/app`
- `go test -race ./...`
- `golangci-lint run ./...`
- `gofmt -l pkg cmd`
- `git diff --check`

## AI usage

- Implemented with GitHub Copilot (Sol 5.6 / Claude Fable)
- Reviewed with Claude Code (Fable 5.1)


pkg/ghinstall/ghinstall.go
```diff
@@ -413,18 +413,20 @@ func (rr *roundRobin) enumerate(
...
 			// Bail rather than walking the remaining managers only to
 			// collect N copies of the same cancellation error.
-			errs = append(errs, ctx.Err())
 			break
 		}
 
 		installs, err := each(m, ctx, owner)
 		if err != nil {
-			errs = append(errs, err)
+			if ctxErr = ctx.Err(); ctxErr == nil {
```
jmeridth: :robot: (low, non-blocking) The inner guard that keeps the canceled manager's own error out of `errs` isn't really pinned. If it went back to always appending while still setting `ctxErr`, both the mid-walk and genuine-failure subtests still pass, because the code comes out Canceled either way and only the warning count changes. Would it be worth asserting the log says `1 of 2 managers failed` in the genuine-failure case, and has no warning in the mid-walk case?

jmeridth (dismissed):
:robot: (summary) Looks correct to me. Every caller of GetAll/GetAllFresh already refuses to trust a partial list when the error is set, so switching to Canceled/DeadlineExceeded doesn't change any absence or caching decision, and the pinConfirmTimeout remap only fires for the walk's own timeout. One small test-coverage note inline, not blocking.

pkg/ghinstall/ghinstall.go
phelian: @jmeridth I agree with the bot. The status assertions didn’t pin whether the canceled manager’s error was counted as a failure. I’ve added a no-warning assertion for mid-walk cancellation and a `1 of 2 managers` failed assertion when a genuine failure precedes cancellation. The focused test passes with -race.

jmeridth: @phelian Thanks, that pins it. Both mutations of the guard now fail the mid-walk and genuine-failure subtests.

jmeridth (approved):
:robot: (summary) Re-approving at 0a9430b. The follow-up is test-only and closes the guard coverage gap; production code is unchanged from the earlier approval.

PR #1694, opened 2026-09-22
Share the clog capture test helper

phelian:
## Summary

- add `internal/logtest.Capture`, which returns a context whose `clog` logger writes to a `bytes.Buffer`
- use it in the `ghinstall` and `octosts` tests that assert on log output, replacing three copies of the same `clog.WithLogger(t.Context(), clog.New(slog.NewTextHandler(&logs, nil)))` setup

Stacked on #1693: the first commit here is that PR's commit and this one only adds the second. Once #1693 merges this rebases to a single commit.

Review only the second commit: https://github.com/octo-sts/app/pull/1694/commits/11d72ca

## Why

The same log-capture incantation now appears in two packages, and the next test that needs to assert on a warning would copy a fourth. A small shared helper keeps the setup in one place and drops the `log/slog` and `clog` imports from the test files.

## Validation

- `go test -race ./pkg/ghinstall/ ./pkg/octosts/`
- `golangci-lint run ./internal/... ./pkg/ghinstall/... ./pkg/octosts/...`
- `gofmt -l internal pkg`
- `git diff --check`

## AI usage

- Implemented with Claude Code (Fable 5.1)


jmeridth (approved):
:robot: (summary) Clean extraction; the helper body is unchanged and the converted tests assert the same things. Looks good.

